PR #142, opened 2022-11-30
Update links in README

shehzan10:


kring:
I only just realized now that this PR accidentally reverted a change to cesium-native. Surprisingly, I can't reproduce the bug that this cesium-native change fixed. I'm investigating, but we may need to do another release.

PR #149, opened 2022-12-05
Add support for Nullable<T> and `?` notation to Reinterop

### LINKED ISSUE #139: Add support for nullables to Reinterop
It would be nice to add support for nullables in Reinterop, e.g. ```public partial double3? ReturnDouble3Nullable();```

kring:
Fixes #139

It works well for methods and nullables of blittable value types. Other scenarios are not as well tested, so let me know if you run into trouble @j9liu.

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -14,7 +14,7 @@ jobs:
           find native~/Runtime native~/Editor native~/Shared \( -iname '*.cpp' -o -iname '*.h' \) -print0 | xargs -0 clang-format --dry-run -Werror
   Windows:
     needs: [QuickChecks]
-    runs-on: windows-latest
+    runs-on: ["self-hosted","windows","x64","unity-2021-3"]
     # Only allow a single Windows build at a time, for Unity licensing reasons
     concurrency: windows
     steps:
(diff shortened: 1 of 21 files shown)


j9liu:
I'm trying this out on my branch but I get the following error:

```
C:\Dev\cesium-unity-samples\Packages\com.cesium.unity\native~\Runtime\generated-Editor\include\DotNet/System/Nullable1.h(5,10): 
fatal error C1083: 
Cannot open include file: 'T.h': No such file or directory [C:\Dev\cesium-unity-samples\Packages\com.cesium.unity\native~\build\Runtime\CesiumForUnityNative-Runtime.vcxproj]
```
It seems like `Nullable` is trying to reference a `T.h` file:

```
#include <T.h>
namespace DotNet::System {

template <typename T0>
struct Nullable1;

template <> struct Nullable1<T> {
  private: friend std::uint8_t (::initializeReinterop)(std::uint64_t validationHash, void** functionPointers, std::int32_t count);
  private: T value;
  private: ::std::int32_t _dummyPrimitive;
};
```

Even when I put something like `double3? nullable = null` in Reinterop to make it generate a definition for `double3`, the `T` definition is still there.

kring:
@j9liu I've added support for partial methods with nullables now, too. So you can add a method like this to a C# class:

```c#
public static partial Unity.Mathematics.double3? TestNullable(Unity.Mathematics.double3? maybeDouble);
```

And that will try to call into a method on the impl class that looks like this:

```c++
static std::optional<DotNet::Unity::Mathematics::double3> TestNullable(const std::optional<DotNet::Unity::Mathematics::double3>& maybeDouble);
```

kring:
Thanks @j9liu, I've addressed those comments.

j9liu:
Looks good, thanks @kring !

Issue #132, opened 2022-11-29
Tileset displacement incorrect after origin change 


nithinp7:
#65 was partially fixed in #122, but still seems to have the following issue.

Currently transforms for tiles in Unity generally get constructed like this:
```tileLocalToUnity = UnityAdjustmentTransform * ECEFToUnity * tileLocalToECEF```
where "UnityAdjustmentTransform" is the hand-edited transform applied in addition to the default tileset transform.

The problem is that the ECEFToUnity transform changes every time the origin moves. This means that a "UnityAdjustmentTransform" relative to an original georeference origin, will be _incorrect_ when composed onto the ECEFToUnity corresponding to a new georeference origin. To fix this, we'll need to do a change of basis on the UnityAdjustmentTransform _at the time the change is commited_, so it becomes an ECEFAdjustmentTransform:
```ECEFAdjustmentTransform = UnityToECEF * UnityAdjustmentTransform * ECEFToUnity```

Later we can compute tile transforms by:
```tileLocalToUnity = ECEFToUnity * ECEFAdjustmentTransform * tileLocalToECEF```
Note the new adjustment matrix is no longer relative to a transient frame-of-reference. Further Unity adjustments can similarly be changed to an ECEF basis and composed with the previous adjustment.

The only tricky part will be automatically detecting / overriding Unity tileset transform events (whether in editor or during play). But I suspect some of these details have already been sorted out over the course of #122 and the globe anchor implementation.

PR #150, opened 2022-12-05
Update georeference when placing origin

nithinp7:
The georeference wasn't being updated during `PlaceGeoreferenceHere`, so the globe-relative camera transform after the change was not being properly restored (especially noticeable when changing the georeference beyond the horizon). Previously this issue had not surfaced since the camera always ended up at Unity's origin - so it did not need the up-to-date georeference transform to compute the new position.

### DIFF
--- a/Editor/CesiumEditorUtility.cs
+++ b/Editor/CesiumEditorUtility.cs
@@ -299,6 +299,8 @@ namespace CesiumForUnity
                 positionECEF.y,
                 positionECEF.z);
 
+            georeference.UpdateOrigin();
+
             double3 newCameraForwardUnity =
                 georeference.TransformEarthCenteredEarthFixedDirectionToUnity(cameraForwardEcef);
 


kring:
@nithinp7 `SetOriginEarthCenteredEarthFixed`, which is called on the line just before the new `UpdateOrigin` call, sets the `originAuthority` property. And that property's setter calls `UpdateOrigin`. So `UpdateOrigin` was already being called before this PR. Is there some bug that is fixed by calling it _twice_? If so, that warrants some more investigation.

nithinp7:
Hm I see, you're right. I'm also not able to reproduce the original issue after reverting this change. And it looks like no one else is running into the issue so maybe this was a false alarm. I wonder if I had the tilesets offset somehow, maybe what I was seeing was #132 - but I'm not sure how this change even gave the illusion of fixing it! Closing this.

PR #154, opened 2022-12-08
Add `CesiumCameraController`

j9liu:
This PR adds the `CesiumCameraController` class, which intends to mimic the `DynamicPawn` in Unreal Engine.

**TODO:**
- [x] Get dynamic speed adjustment working (most of the blueprint from Unreal is copied over but it's still not working) 
- [x] Dynamically adjust near clip plane based on height from ellipsoid
- [x] Polish user UI / API
- [x] Add "Dynamic Pawn" option to Cesium QuickAdd


j9liu:
While working on dynamic clipping planes for the camera controller, I found that if the camera's far plane is too large (usually a value over `10,000,000` Unity will start printing these errors:

image

I can't pinpoint why these happen -- when I look up the error it seems like users can get this error despite different camera setups. Might have to put a pin in this and look at it later.

j9liu:
I think this PR is ready for review! It can be tested with the new sample scene in https://github.com/CesiumGS/cesium-unity-samples/pull/31. Though this probably should wait until #149 is merged.

j9liu:
> While working on dynamic clipping planes for the camera controller, I found that if the camera's far plane is too large (usually a value over `10,000,000` Unity will start printing these errors:

It seems like a near clip of 500,000 or more results in this error too, so I capped it at 450,000 and doing so alleviated the range in which the errors occur. The errors still happen if you zoom out far enough, but it takes a little longer to get there.

j9liu:
More clipping planes shenanigans: based on our offline conversation I tried two different approaches today.

1. Mimic what CesiumJS does by testing for all objects visible from the camera (via their bounding volumes), then adjusting the near / far clips based on what bounding volumes are on-camera.
   - I actually got this to work _somewhat_ well, though I do think it ends up somewhat hacky because the offset of the near-clip plane seems hard to get right. It makes the near plane nearer but it often clips into the nearby objects, so they do show up, but they're partially cut off (e.g. cube is sliced through its corner).
   - This method will also potentially result in the globe flickering in and out of visibility, if the camera is barely seeing an object close to the camera, and then moves to the side to look at only the globe. Though that would have to be a significant altitude above the earth, so maybe that's what we'd have to settle for with this method.
   - Unfortunately, if this method needs to account for 3D tiles, we'll have to use `Resources.FindObjectsOfTypeAll`, which is documented as an expensive function that shouldn't be called every frame. My performance didn't seem that bad from above the Earth but this does iterate over all hidden objects, including inactive tileset meshes. If we did this we'd probably have to throttle it by only doing the operation every X frames.
   - Alternatively, we can just do the regular `FindObjectsOfType` and only look for user-placed objects that are active + have mesh renderers. Then we would only do near-clip adjustment.
2. Use multiple cameras, one to render objects up close and one to render the faraway globe.
  - This requires using camera layers like https://www.youtube.com/watch?v=bbnVpPiQ_rU
  - Basically objects up close would need to be put on a separate camera layer, to render it on top of the output from the camera for the globe. Again, need to iterate over renderers using `FindObjectsOfType` but since we'd only be checking for nearby, user-inserted elements I don't think this will be that bad.
  - I still have to get it to work because when I did it without camera layers, it would only render one output or the other, and I found this method JUST before leaving work.

j9liu:
Well... I'm pretty close to giving up on the dynamic clipping planes.

Tried to multiple cameras approach today, and unfortunately accomplishing the same effect in all three render pipelines requires different API and camera settings.

- For Unity's built-in renderer this is really easy, just need to 
- For URP, the camera to render the globe far away has to be the "Base" camera, while the camera for rendering nearby objects has to be the "Overlay" camera.
- For HDRP, haven't figured out a good way to do this without using the Graphics Compositor (which is REALLY discouraged for games because it's not performant). There's also no good way to make a camera render to a transparent texture, so I can't get the cameras to overlay each other easily without some shader code. I found a [suggestion to create a Custom Pass Volume](https://www.reddit.com/r/Unity3D/comments/ls4e6a/depth_only_option_in_hdrp/) that clears the output of a camera but I haven't been able to try it. HDRP seems to be the most difficult to get this to work.

The second camera's component would need to be modified at runtime based on the three different pipelines. However, the URP and HDRP API are not immediately available without dependencies on the relevant packages. So I put those packages as dependencies in the project and added them to the `asmdef`, which worked. _However_, ever since I did that my Unity Editor was running extremely slowly. Unity kept taking at least 20 minutes to recompile / reopen my project so this seems extremely unideal.

Other options:
- Rendering to a render texture and then compositing the output: although this is used by both URP and HDRP I believe this would require shaders to actually add the images together. I don't know if the same shader can be usable by both pipelines.
- Assigning nearby objects to a layer mask, then rendering with a second camera. The problem is you can't create a layer at runtime or by script, so the user would have to create a layer and assign any necessary objects they need to show up with that layer. Ultimately, if a user needs to do additional work like this, it seems better to let them handle it with their own camera setup / layer settings instead.

After assessing this, I think its best to revert to the first method, and just try to fine-tune the clipping plane adjustment behavior. Or just offer the option for users to provide their own values as needed, when it rises above a certain altitude.

j9liu:
@kring this is finally ready for review, when you get the chance!

kring:
I posted a question to the Unity forum about this:
https://forum.unity.com/threads/is-there-a-hard-maximum-far-near-clip-plane-ratio.1377387/

Runtime/CesiumCameraController.cs
```diff
@@ -0,0 +1,1145 @@
...
+
+        private void HandleSpeedChange(float speedChangeInput)
+        {
+            if (this._enableDynamicSpeed)
+            {
+                this.UpdateDynamicSpeed();
+            }
+            else
+            {
+                this.SetMaxSpeed(100.0f);
```
kring, line 511: I think it would be very useful for this to be configurable.

Runtime/CesiumCameraController.cs
```diff
@@ -0,0 +1,1145 @@
...
+        #endregion
+
+        #region Fly-To helper functions
+
+        /// <summary>
+        /// Advance the camera flight based on the given time delta.
+        /// </summary>
+        /// <remarks>
+        /// <para>
+        /// This function requires the CesiumGlobeAnchor to be valid. If it is not valid,
```
kring, line 728: Do you mean the `CesiumGeoreference`? I don't think the CesiumGlobeAnchor can be missing, because of the `RequireComponent` attribute.

Runtime/CesiumCameraController.cs
```diff
@@ -0,0 +1,1145 @@
...
+            {
+                return;
+            }
+
+            float nearClipPlane = this._initialNearClipPlane;
+            float farClipPlane = this._initialFarClipPlane;
+
+            if (height >= this._dynamicClippingPlanesMinHeight)
+            {
+                farClipPlane = height + (float)(2.0 * CesiumEllipsoid.GetMaximumRadius());
```
kring, line 1124: Is a 2x max radius necessary? I think anything beyond one radii would be on the back side of the globe anyway, right? Of course if you have stuff you want to see in orbit, then either one might be insufficient, but there's probably not a whole lot we can do about that given Unity's severe clipping plane limitations.

Runtime/CesiumCameraController.cs
```diff
@@ -0,0 +1,1145 @@
...
+
+                float farClipRatio = farClipPlane / this._maximumNearToFarRatio;
+
+                if (farClipRatio > nearClipPlane)
+                {
+                    nearClipPlane = Mathf.Min(farClipRatio, this._maximumNearClipPlane);
+                }
+                else
+                {
+                    nearClipPlane = 10.0f;
```
kring, line 1135: Is this right? Why 10.0f rather than whatever the initial near plane was? If the initial near plane happened to be large, we could end up in this else block even though a near plane of 10.0f would cause Unity to cut off things in the distance because of its 100,000:1 far:near requirement.

kring (commented):
This looks really great @j9liu! Just a few questions/suggestions for you below.

Runtime/CesiumCameraController.cs
```diff
@@ -0,0 +1,1145 @@
...
+        private double _flyToGranularityDegrees = 0.01;
+
+        /// <summary>
+        /// The granularity in degrees with which keypoints should be generated for the 
+        /// flight interpolation.
+        /// </summary>
+        public double flyToGranularityDegrees
+        {
+            get => this._flyToGranularityDegrees;
+            set => this._flyToGranularityDegrees = Math.Max(value, 0.0);
```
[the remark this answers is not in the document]
j9liu: Looks like it does this in Unreal too :grimacing: probably because the code doesn't check for division by zero.

Runtime/CesiumCameraController.cs
```diff
@@ -0,0 +1,1145 @@
...
+        /// <summary>
+        /// A curve that dictates what percentage of the max altitude the camera 
+        /// should take at a given time on the curve.
+        /// </summary>
+        /// <remarks>
+        /// This curve must be kept in the 0 to 1 range on both axes. The 
+        /// <see cref="CesiumCameraController.flyToMaximumAltitudeCurve"/>
+        /// dictates the actual max altitude at each point along the curve.
+        /// </remarks>
+        public AnimationCurve flyToAltitudeProfileCurve
```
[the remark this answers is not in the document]
j9liu: That makes sense. I'll separate the code into a component named `CesiumFlyToController`, but feel free to suggest other names!

Runtime/CesiumCameraController.cs
j9liu: Ah thanks for the catch. This is leftover from when I was trying out different approaches / values to make it work; the code here got messy

Runtime/CesiumCameraController.cs
j9liu: Yeah I thought that one radii would be enough, but when I did it with just one radii, the edge of the Earth was noticably "bouncing" as it zoomed out.
OnlyRadius

So I multiplied it by 2 so that the bouncing wouldn't happen at that a distance where the Earth was still relatively sizeable on the screen.

TwiceRadius

Of course, once it gets extremely small we run into the same effect, but it takes longer to get there.
 
TwiceRadiusVeryFarAway

If there's a more robust way of calculating this scalar, or if it's better to leave it as one radii and just communicate how to set up rendering environments for objects that are far away, let me know.

Runtime/CesiumCameraController.cs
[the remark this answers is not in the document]
j9liu: During this, I came up with some UX / API questions that I wanted to think through:
- `FlyTo` had an option to make the flight interruptible by movement. It also disabled mouse input to rotate the camera when it was in flight. By separating `CesiumFlyToController` from `CesiumCameraController`, we're separating the flight behaviour from the camera movement, so these classes will need to communicate with each other.
- A public `InterruptFlight` method in `CesiumFlyToController` seems sufficient; in their own scripts, a developer can manage if / when the flights can be interrupted. But what about the other way around in `CesiumCameraController`? Or if the flight shouldn't be interrupted at all, how do you disable the input altogether?
  - Perhaps this isn't a problem since Fly-to is manually imposing positions / rotations on the camera, so it theoretically overrides it. But I wasn't sure if it was good form to _not_ disable inputs on `CesiumCameraController`.

This made me question the best user interface for `CesiumCameraController`, once `CesiumFlyToController` becomes its own component.
- Is it helpful for `CesiumCameraController` to have an interface similar to `CharacterController` where users can move the camera by script? So for example, `MoveRight()`, `MoveUp()`, `Rotate()` etc. functions. Users can write their own scripts using those functions if they need.
- But then how would you change / use / disable key bindings? Currently they're baked into `CesiumCameraController`, and I don't think we should remove them because some people won't want to write their own controllers.

So I suppose it is a matter of potential use cases, if we ever a user will want to disable or override the camera bindings. I'm probably overthinking it, but I just wanted to throw this out there before I reorganize all of this code, just in case.

I looked around some free controllers on the Asset Store, and they all do different things. One of them was an interface that you needed to inherit from, one of them had key bindings / controls baked into the script, and one of them allowed users to modify the controls via the Inspector. So it's a matter of deciding how we want to treat our controller classes based on what our users would want.

Runtime/CesiumCameraController.cs
kring: Ok, I played around with this and... I have no idea what Unity is doing. But sticking to 2x for now is fine.

Runtime/CesiumCameraController.cs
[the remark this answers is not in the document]
kring: How's this sound?

* `CesiumFlyToController` knows about and uses `CesiumCameraController`, but not the other way around. In `Awake`, `CesiumFlyToController` looks for a `CesiumCamerController` and will use it if present, but it is not required.
* Add `EnableRotation` and `EnableMovement` properties to `CesiumCameraController`.
* At the start of a flight, `CesiumFlyToController` sets `EnableRotation` to false. If flight interruption is not allowed, it also sets `EnableMovement` to false (otherwise it leaves it true).
* If a flight may be interrupted, `CesiumFlyToController` checks at each step if the position is different from what it previously set in the last step (in ECEF probably). If so, the flight is interrupted. So the `CesiumCameraController` doesn't need to know anything about flights, it just moves as normal (if movement is enabled) and this causes the flight to be interrupted. This allows flights of other random objects to get interrupted when the objects are moved explicitly.
* At the end of the flight, `EnableRotation` and `EnableMovement` are restored to their original values.

> Is it helpful for CesiumCameraController to have an interface similar to CharacterController where users can move the camera by script? So for example, MoveRight(), MoveUp(), Rotate() etc. functions. Users can write their own scripts using those functions if they need.

This sounds useful. Maybe even `MoveNorth` etc. But I think it can be a future feature because introducing it later doesn't have any backward compatibility implications.

> But then how would you change / use / disable key bindings? 

I believe in Unreal there's an abstraction here, where you can define conceptual inputs (like "move left"), and then separately and via runtime configuration bind those inputs to specific keys / controls / actions. I guess there's nothing like that in Unity? If not, I think we ultimately should allow the user to configure the bindings using the Inspector and by setting properties at runtime, but that can be a future feature (again, no backward compatibility concerns really).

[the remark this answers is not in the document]
j9liu: @kring This all sounds reasonable to me. Thanks for thinking it through with so much detail :pray:

j9liu:
@kring updated based on your feedback!

kring:
I updated the changelog to mention the rename of CesiumTransforms as a breaking change. And changed the next version to be 0.2.0 instead of 0.1.3. Otherwise, looks great, merging!

PR #158, opened 2022-12-15
convert 'X-Cesium-Platform' to the rfc8187 format

### LINKED ISSUE #155: Cesium tries to add invalid characters to HTTP request headers
Probably when the project name uses non-ASCII characters, which Unity (and probably HTTP in general) do not allow.

As reported here:
https://community.cesium.com/t/launching-cesium-for-unity/21380/12

We can likely fix this by encoding as described in [RFC 8187](https://www.rfc-editor.org/rfc/rfc8187).

joseph-kaile:
Fixes #155 

Attempted to follow the format of https://www.rfc-editor.org/rfc/rfc8187. It adds utf-8 to the beginning, replaces spaces with %20 (assumes the spaces are unquoted), and replaces unicode character with the hex format.

image

### DIFF
--- a/native~/Shared/src/UnityAssetAccessor.cpp
+++ b/native~/Shared/src/UnityAssetAccessor.cpp
@@ -23,9 +23,7 @@
 #include <DotNet/UnityEngine/Networking/UploadHandler.h>
 #include <DotNet/UnityEngine/Networking/UploadHandlerRaw.h>
 
-#include <fstream>
 #include <sstream>
-
 
 using namespace CesiumAsync;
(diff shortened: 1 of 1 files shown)


kring:
Thanks @joseph-kaile, this looks good. I want to hold off merging it until @mramato has a chance to weigh in, though, because it has implications for the ability to parse information from the logs. Can I bug you to implement a simpler version of this in a separate PR that just replaces invalid characters with question marks? That allows us to fix the immediate problem without totally changing the format.

PR #159, opened 2022-12-16
Use self-hosted runner to build the package

kring:


kring:
Windows builds now take about half as long as before, and should cost less as well. Also fixed a bug causing the build/package to spin wait while Unity was running, because `StreamReader.ReadLine` doesn't block while waiting for more input like I thought it did.

PR #160, opened 2022-12-20
replace invalid characters with question marks

joseph-kaile:


kring:
Looks great, thanks @joseph-kaile!

Issue #104, opened 2022-11-18
`CesiumGeoreference`, `CesiumSubScene`, `CesiumGlobeAnchor` don't refresh with pasted values


j9liu:
Unity has the option to copy the values of a component and paste them into another component's fields.

<img src="https://user-images.githubusercontent.com/32226860/202779366-1620f4f2-f9e3-47f5-a44b-0542fd6aeca3.png" width="300px"/>

 However, the aforementioned components don't automatically refresh when values are pasted into them. I think the paste bypasses any checks done by the custom editor, and just changes the `SerializedObject` underneath. This means our components won't know to update themselves unless the values are edited manually in the editor fields themselves.

I'm not sure it's an easy fix to do some code in `OnValidate`, because `OnValidate` triggers whenever the custom editors change a value as well. Plus, `OnValidate` can't tell what property / properties changed. So the check might have to be in the custom editors. Before `serializedObject.Update()` is called, you may have to store the old values of the `serializedObject`, then compare them at the end of `OnInspectorGUI`. Then, if things have changed, refresh the respective components.

PR #163, opened 2022-12-20
Add Undo support for `CesiumGeoreference`, `CesiumSubScene`, and `CesiumGlobeAnchor`

### LINKED ISSUE #117: Implement georeference origin change "undo" in editor
In the editor, it would be useful to be able to Ctrl-Z undo a georeference origin change.

j9liu:
Fixes #117 and #104.


j9liu:
Ended up using `OnValidate` for all of the classes. I think it's the most elegant solution and as long as these classes aren't handling object creation ([as noted in the documentation](https://docs.unity3d.com/ScriptReference/MonoBehaviour.OnValidate.html)), it should be fine. This now fixes #104 as well.

j9liu:
@kring It looks like calling `Sync()` in `CesiumGlobeAnchor.OnValidate` is causing CI to fail because of the exception that it's not nested under a `CesiumGeoreference`. What's the best course of action -- should I put that line of code in a `try` block?

Editor/CesiumEditorUtility.cs
```diff
@@ -325,7 +325,7 @@ public static CesiumSubScene CreateSubScene(CesiumGeoreference georeference)
                 georeference.ecefZ
             );
 
-            GameObject subSceneGameObject = new GameObject();
+            GameObject subSceneGameObject = new GameObject("New Sub-Scene");
```
j9liu, line 328: Small unrelated change that is in this PR for convenience.

kring:
@j9liu this is causing a test failure in `SyncsUnityPositionFromTransformAndBackMultipleFrames` in `TestCesiumGlobeAnchor.cs`. (note to self: really need to get these tests, such as they are, running automatically on CI).

It seems that `OnValidate` is called immediately from `AddComponent`, so we can no longer observe the positionAuthority as "None" after adding the component. That's not a problem by itself, but it means there will be a difference in how the CesiumGlobeAnchor works in the Editor and in the Player:

* _In the player_: Adding the component does not immediately invoke Sync. So you can add the component and then immediately set the position, and this won't affect the orientation of the object.
* _In the editor_: Adding the component calls OnValidate, which "locks in" the position. If you then move it to the other side of the globe, its orientation will be flipped.

I'm not sure how big a problem this is. But fundamentally, the GlobeAnchor state transitions are ... really hard to get right. And it's even harder to be sure they're right. And I think letting a "called all the time, for lots of different reasons" method like OnValidate affect the GlobeAnchor's state makes it even harder to reason about.

So... is there a finer-grained alternative to OnValidate that would solve the initial problem here? If not, I'm sure we can make it work, but I think it's worth thinking about any alternatives.

kring:
I added a new test for the scenario I described above. It's failing in this branch, but if you remove OnValidate, it passes.

j9liu:
So while debugging, it looks like doing `OnValidate()` was wrong even for `CesiumGeoreference`, because when a script loads it calls `OnValidate` before `OnEnable`, so it was calling `UpdateOrigin` before `InitializeOrigin` :face_palm: It was actually causing the globe anchors to be disoriented, even without `OnvValidate()` in the globe anchor's code.

I might just give up and make all of these classes cache values that they compare their current values against. I have to do that for `CesiumGlobeAnchor` anyway and it would be nice to make it consistent across all related classes.

kring:
This looks great, @j9liu. I like that only the Editor components have to deal with what amounts to an Editor-only problem, even if it's a bit of boilerplate.

PR #169, opened 2023-01-08
Build ios

joseph-kaile:
Switching platform to iOS, then building will generate an XCode project. Opening the XCode project, you will need to set up the provisioning profile. After that it should be able to install on an iphone/ipad without any other issues.

This requires cesium-native with the sprintf fix from the main branch in order to work.


Editor/CompileCesiumForUnityNative.cs
```diff
@@ -216,6 +221,23 @@ public void OnPostBuildPlayerScriptDLLs(BuildReport report)
...
+        public void OnPostprocessBuild(BuildReport report)
+        {
+#if UNITY_IOS
+            if(report.summary.platform == BuildTarget.iOS)
+            {
+                string projectPath = report.summary.outputPath + "/Unity-iPhone.xcodeproj/project.pbxproj";
+                PBXProject pbxProject = new PBXProject();
+                pbxProject.ReadFromFile(projectPath);
+                string target = pbxProject.GetUnityMainTargetGuid();
+                pbxProject.SetBuildProperty(target, "ENABLE_BITCODE", "NO");
```
kring, line 234: Should we be building our native code with bitcode enabled, rather than disabling it here?

Editor/CompileCesiumForUnityNative.cs
```diff
@@ -255,6 +277,17 @@ public static LibraryToBuild GetLibraryToBuild(PlatformToBuild platform, string
             if (platform.platformGroup == BuildTargetGroup.Android)
                 library.Toolchain = "extern/android-toolchain.cmake";
 
+            if (platform.platformGroup == BuildTargetGroup.iOS){
+                var outputPath = Path.Combine(library.SourceDirectory, "../Plugins/iOS");
+            if (Directory.Exists(outputPath))
+                Directory.Delete(outputPath, true);
+            Directory.CreateDirectory(outputPath);
```
kring, line 284: Are these lines needed, or were they just for testing? If they're needed, fix the indentation and maybe add a comment explaining why.

Editor/CompileCesiumForUnityNative.cs
```diff
@@ -284,6 +321,8 @@ private static string GetDirectoryNameForPlatform(BuildTargetGroup platformGroup
         {
             if (IsEditor(platformGroup, platform))
                 return "Editor";
+            else if (IsIOS(platformGroup, platform))
+                return "iOS";
```
kring, line 325: Is this needed? If I'm not mistaken, `platformGroup.ToString()` will return `"iOS"` as well.

Editor/CompileCesiumForUnityNative.cs
```diff
@@ -358,6 +397,9 @@ internal static void BuildNativeLibrary(LibraryToBuild library)
                     args.AddRange(library.ExtraBuildArgs);
                     startInfo.Arguments = string.Join(' ', args);
                     RunAndLog(startInfo, log, logFilename);
+                 
+                    if(library.Platform == BuildTarget.iOS)
+                        AssetDatabase.Refresh();
```
kring, line 402: Why is this needed on iOS specifically?

kring (commented):
Looks good @joseph-kaile! A few questions/suggestions below.

Editor/CompileCesiumForUnityNative.cs
```diff
@@ -255,6 +255,20 @@ public static LibraryToBuild GetLibraryToBuild(PlatformToBuild platform, string
...
+
+            if(platform.isCleanBuild)
+                Directory.Delete(outputPath, true);
+
+            if (!Directory.Exists(outputPath))
+                Directory.CreateDirectory(outputPath);
+
+                library.Toolchain = "extern/ios-toolchain.cmake";
+                library.ExtraConfigureArgs.Add("-GXcode");
+                library.ExtraConfigureArgs.Add("-DCMAKE_INSTALL_PREFIX=../Plugins/iOS");
```
kring, line 269: I think these lines are inside the wrong `if`.

Editor/CompileCesiumForUnityNative.cs
kring: Joseph pointed out offline that it actually returns `iPhone` because there are multiple enums with the same value. Resolving.

Editor/CompileCesiumForUnityNative.cs
kring: We discussed offline that this avoids the need to create placeholders for all the .a files. We might be able to do the same on all the other platforms, too, and save some trouble. It's fine for now, though.

kring (commented):
Looking great, I think just that misplaced code in the wrong conditional needs to be fixed and then I'll merge this.

Editor/CompileCesiumForUnityNative.cs
joseph-kaile: Figured out that these lines were redundant.

joseph-kaile:
Had to revert removing placeholder code because it didn't work. After reverting, I tested that built and pre-built editor/build/ios works.

kring:
Thanks @joseph-kaile!

PR #174, opened 2023-01-11
Touch controller

joseph-kaile:
Adds touch control support to the dynamic camera controller.


Runtime/CesiumCameraController.cs
```diff
@@ -7,6 +7,10 @@
 using UnityEngine.InputSystem;
 #endif
 
+#if UNITY_IOS || UNITY_ANDROID
+using Touch = UnityEngine.InputSystem.EnhancedTouch;
```
j9liu, line 11: Can this be called `EnhancedTouch` instead? Looking at something like `Touch.Touch.activeTouches` is confusing

Runtime/CesiumCameraController.cs
```diff
@@ -449,11 +456,39 @@ private bool RaycastAlongForwardVector(float raycastDistance, out float hitDista
...
+            lookDelta = lookAction.ReadValue<Vector2>();
+            moveDelta = moveAction.ReadValue<Vector2>();
+
+#if UNITY_IOS || UNITY_ANDROID
+            bool handledMove = false;
+            bool handledLook = false;
+
+            foreach(var touch in Touch.Touch.activeTouches)
+            {
+                if(touch.startScreenPosition.x < Screen.width / 2)
```
j9liu, line 470: I don't have the ability to test this on my own, so do you mind clarifying what the controls are? I see that you can move by dragging on the left hand side of the screen, but what does `touch.delta` mean?

j9liu: Also I assume this is without some sort of UI, right? So the user just drags on the screen, on top of their scene?

j9liu (commented):
Hey @joseph-kaile,

I just took a look at your changes. The demo you showed offline looks pretty good, though I wish I had the ability to test it myself. I've also never navigated a 3D app on my phone before; is this the standard way of navigating in 3D on mobile?

Also, I assume a bunch of these changes are from #169, so I only looked at `CesiumCameraController.cs`. Let me know if I'm wrong.

Runtime/CesiumCameraController.cs
joseph-kaile: Yes, thanks that is better.

Runtime/CesiumCameraController.cs
joseph-kaile: Yep, usually you play mobile games with thumbs in landscape mode. the left thumb moves the player, and the right thumb moves around. There is usually a circle with a dot inside that move like thumb-sticks. Here there is no UI, however I don't think its necessary to show this.

The left thumb doesn't use delta, so that the camera moves proportional to the amount it moved from the starting position. This way the camera still moves even without moving the thumb.

The right thumb uses delta so that the camera only looks around when the thumb moves. The sensitivity for both thumbs works out perfectly so it doesn't need to be scaled.

j9liu: That makes sense, thanks for explaining :pray:

kring:
@joseph-kaile can you please merge in main to fix the conflicts?
@j9liu any reservations about shipping this in February's release?

j9liu:
@kring No, this can go in the Feburary release. My only reservation came from whether or not users could infer these mobile controls, or if we had to tell them / use UI somehow. But @joseph-kaile says these are standard controls, so I'll take his word for it.

j9liu:
@joseph-kaile do you mind re-merging with main so CI can pass + I can merge this PR?

j9liu:
Thanks @joseph-kaile !

PR #183, opened 2023-01-19
Change specified folder of `CesiumRuntimeSettings.asset`

j9liu:
Previously, `CesiumRuntimeSettings` would create / move itself under `Assets/Settings/Resources`. This changes the file path to `Assets/CesiumSettings/Resources` so that `CesiumRuntimeSettings` is separated from other settings in the user's project.


argallegos (approved):
Looks good and functions as expected. Thanks, @j9liu

PR #185, opened 2023-01-24
Use `OnValidate` to prevent constant editor changes to `CesiumGlobeAnchor`

j9liu:
#163 unfortunately created a bug with `CesiumFlyToController` when the `DynamicCamera` is selected in the Editor. Because `CesiumGlobeAnchor` constantly changes its values, the editor misinterprets the changes as user input. It constantly recomputes its coordinates, leading to small changes from the previous frame, which the `CesiumFlyToController` interprets as user inputted movement. The changes to the `CesiumGlobeAnchor` while in flight will be incorrectly interpreted as user movement, and the flight will be interrupted.

We can't reliably distinguish the source of changes to a `SerializedObject` without `OnValidate`. In the past, I thought to check for some sort of "initialization" variable in `OnValidate` before calling other update functions. However, when I last tried it, the updates would still go through despite the environment reloading. I spent some time looking into why this happened, and I found:
- `OnValidate` in a `Monobehaviour` is reliably called before its `OnEnable` (though not necessarily before the `OnEnable` of other classes).
- When you load a new scene, private variables (such as `_lastPropertiesAreValid` in `CesiumGlobeAnchor`) are set to their initial values. This means that they will be properly "initialized" in `OnEnable` and won't incorrectly update in `OnValidate`.
- When you either recompile scripts or switch to play mode, the private variables are **not** reset to their initial values. `OnValidate` is still called before `OnEnable`, and this causes things to get wonky in between what you'd perceive as complete environment reloads. Perhaps this is because of `[ExecuteInEditMode`]; they seem to reuse the same `Monobehaviour` between these states, so the variables don't get reloaded. 
- `OnDisable` is always called before an environment switch. So the order of operations for a `Monobehaviour` in a scene is reliably `OnDisable` -> `OnValidate` -> `OnEnable`.

With this, I revisisted the `OnValidate` approach. The relevant functions are only called if the classes have been "initialized." This "initialization" state is reset whenever the object is disabled, so it shouldn't matter if it's the same `Monobehaviour` between environment reloads; it'll always reset. A summary of what I did to each class:
1. `CesiumGlobeAnchor`: I use `_lastPropertiesAreValid` as the "initialization" variable. This should prevent unnecessary changes as it's transitioning between the different Unity environments. I saw these incorrect changes happen to `DynamicCamera` because its z-rotation (roll) would get set to a non-zero value. After this PR, the roll stays zero when the scene is loaded, and is always reset to zero after play mode.
2. `CesiumGeoreference`: I added an `_initialized` variable to track when `InitializeOrigin` was called. `UpdateOrigin` will only call `RecalculateOrigin` if the origin was already initialized.
3. 'CesiumSubScene`: Calling `UpdateOrigin` has no consequences for the sub-scene itself, but it sometimes messes up the `CesiumGeoreference`. When sub-scenes get enabled before the georeference, they're calling `UpdateOrigin` on the georeference before it's initialized. But this should be addressed by the changes to `CesiumGeoreference` itself, so no additional work is needed in `CesiumSubScene`.

### DIFF
--- a/Editor/CesiumGeoreferenceEditor.cs
+++ b/Editor/CesiumGeoreferenceEditor.cs
@@ -29,20 +29,6 @@ namespace CesiumForUnity
         private SerializedProperty _ecefY;
         private SerializedProperty _ecefZ;
 
-        // These variables store the values from the previous frame. This is used to compare
-        // to the current serialized values, whether they were changed this frame via user input
-        // or by other means (undo, pasting component values).
-        private CesiumGeoreferenceOriginAuthority
-            _previousOriginAuthority = CesiumGeoreferenceOriginAuthority.LongitudeLatitudeHeight;
-
-        private double _previousLatitude = 0.0;
-        private double _previousLongitude = 0.0;
-        private double _previousHeight = 0.0;
-
-        private double _previousEcefX = 0.0;
-        private double _previousEcefY = 0.0;
-        private double _previousEcefZ = 0.0;
-
         private void OnEnable()
         {
             this._georeference = (CesiumGeoreference)this.target;
@@ -57,8 +43,6 @@ namespace CesiumForUnity
             this._ecefX = this.serializedObject.FindProperty("_ecefX");
             this._ecefY = this.serializedObject.FindProperty("_ecefY");
             this._ecefZ = this.serializedObject.FindProperty("_ecefZ");
-
-            this.UpdateSavedPropertyValues();
         }
 
         public override void OnInspectorGUI()
@@ -75,9 +59,6 @@ namespace CesiumForUnity
             this.DrawEarthCenteredEarthFixedProperties();
 
             this.serializedObject.ApplyModifiedProperties();
-
-            this.UpdateGeoreference();
-            this.UpdateSavedPropertyValues();
         }
 
         private void DrawInspectorButtons()
@@ -203,48 +184,6 @@ namespace CesiumForUnity
             EditorGUILayout.PropertyField(this._ecefZ, ecefZContent);
 
             EditorGUI.EndDisabledGroup();
-        }
-
-        private bool OriginAuthorityChanged()
-        {
-            return this._previousOriginAuthority != this.originAuthority;
-        }
-
-        private bool LongitudeLatitudeHeightChanged()
-        {
-            return this._previousLatitude != this._latitude.doubleValue ||
-                this._previousLongitude != this._longitude.doubleValue ||
-                this._previousHeight != this._height.doubleValue;
-        }
-
-        private bool EarthCenteredEarthFixedChanged()
-        {
-            return this._previousEcefX != this._ecefX.doubleValue ||
-                this._previousEcefY != this._ecefY.doubleValue ||
-                this._previousEcefZ != this._ecefZ.doubleValue;
-        }
-
-        private void UpdateGeoreference()
-        {
-            if (this.OriginAuthorityChanged() ||
-                this.LongitudeLatitudeHeightChanged() ||
-                this.EarthCenteredEarthFixedChanged())
-            {
-                this._georeference.UpdateOrigin();
-            }
-        }
-
-        private void UpdateSavedPropertyValues()
-        {
-            this._previousOriginAuthority = this._georeference.originAuthority;
-
-            this._previousLatitude = this._georeference.latitude;
-            this._previousLongitude = this._georeference.longitude;
-            this._previousHeight = this._georeference.height;
-
-            this._previousEcefX = this._georeference.ecefX;
-            this._previousEcefY = this._georeference.ecefY;
-            this._previousEcefZ = this._georeference.ecefZ;
         }
     }
 }
--- a/Editor/CesiumGlobeAnchorEditor.cs
+++ b/Editor/CesiumGlobeAnchorEditor.cs
@@ -35,23 +35,6 @@ namespace CesiumForUnity
         private SerializedProperty _unityY;
         private SerializedProperty _unityZ;
 
-        private bool _previousDetectTransformChanges = false;
-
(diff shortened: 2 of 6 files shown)


kring:
I've been exploring the lifecycle of MonoBehaviours with `[ExecuteInEditMode]`, so I'm going to write down some notes here for lack of a better place.

It's really poorly documented. For a truly vivid illustration, see [this Unity forum thread](https://forum.unity.com/threads/execution-order-of-events-for-scripts-with-executineditmode-attribute.440047/) where people complain about it, someone at Unity agrees and states their intention to solve this, and then we learn later that he has died. No, I'm not kidding, and sympathies to his family, friends, and colleagues. But it seems possible he was the only person at Unity that really knew how this worked.

So I created a MonoBehaviour with `ExecuteInEditMode` and lots of logging of events. In Unity 2021.3.14, in case it matters. Some things I learned:

* When such a component is added to a game object in the Editor, we get the following sequence of events on "MonoBehaviour A" in this order:
  * Object constructor
  * Awake
  * OnEnable
  * OnValidate
  * Start
  * Update (repeatedly)
* When we hit Play, Unity calls OnDisable on that object and then constructs a brand new object "MonoBehaviour B" and runs it through the sequence:
  * Object constructor
  * OnValidate
  * OnEnable
  * OnDisable
  * OnDestroy
* Yes, MonoBehaviour B is destroyed and thrown away. No idea why. I also love that OnValidate and OnEnable happen in opposite order here. But anyway this object is thrown away, so it's inconsequential I guess.
* After throwing away the object created above, it creates _another_ new object "MonoBehaviour C" and runs through this sequence:
  * Object constructor
  * OnValidate
  * Awake
  * OnEnable
  * Start
  * Update (repeatedly)
* When we stop play, it does the following on MonoBehaviour C:
  * OnDisable
  * OnDestroy
* And then recreates the Editor object (MonoBehaviour D, it doesn't reuse MonoBehaviour A) with the following sequence: 
  * Object constructor
  * OnValidate
  * Awake
  * OnEnable
  * Start
  * Update (repeatedly)

So that's the basic sequence. I was also curious about which field values are preserved when starting Play mode. So I added three fields, one decorated `[SerializeField]`, a second decorated `[NonSerialized]` and a third without any attributes at all. I initialize all three to the value 100 in the constructor, and in `OnEnable`, I increment the value of each and print them.

Long story short, only the private field marked `[SerializeField]` is preserved when entering play mode. Others reset to their constructor-provided value. Presumably public fields would behave the same way, but I didn't test.

Amusingly, the temporary object created at the start of entering play mode (MonoBehaviour B above) also preserves the value of the _undecorated_ field. But I can't see any way it matters.

What about on domain reload, like when we edit a script? That runs through the following sequence:
  * MonoBehaviour A: OnDisable
  * MonoBehaviour E: Object constructor
  * MonoBehaviour E: OnValidate
  * MonoBehaviour E: OnEnable

And here both the `[SerializeField]` and the undecorated field are preserved. A domain reload while Play mode is active looks to be identical, except of course OnDisable happens on the play mode object (MonoBehaviour C) rather then the edit mode one.

One way to understand this is that domain reload doesn't destroy (OnDestroy) and recreate (Awake / Start), but instead just disables and reenables it with some new code.

Here's the code I used to test all this:

```c#
using System;
using System.Runtime.CompilerServices;
using UnityEngine;

[ExecuteInEditMode]
public class LifecycleTester : MonoBehaviour
{
    public LifecycleTester()
    {
        Log("Construct");
    }

    void Awake()
    {
        Log("Awake");
    }

    void Start()
    {
        Log("Start");
    }

    private int _updateCount = 0;
    private int _lastPrintUpdateCount = 0;

    void Update()
    {
        ++this._updateCount;
        int diff = this._updateCount - _lastPrintUpdateCount;
        if (this._updateCount == 1 || diff > 1000)
        {
            Log("Update " + this._updateCount);
            this._lastPrintUpdateCount = this._updateCount;
        }
    }
    //
    private int _undecorated = 100;
    [SerializeField]
    private int _serialized = 100;
    [NonSerialized]
    private int _nonSerialized = 100;

    private void OnEnable()
    {
        Log("OnEnable");

        ++this._undecorated;
        ++this._serialized;
        ++this._nonSerialized;
        Log("OnEnable undecorated=" + this._undecorated);
        Log("OnEnable serialized=" + this._serialized);
        Log("OnEnable nonserialized=" + this._nonSerialized);
    }

    private void OnDisable()
    {
        Log("OnDisable");
    }

    private void OnDestroy()
    {
        Log("OnDestroy");
    }

    private void OnValidate()
    {
        Log("OnValidate");
    }

    private void Log(string text)
    {
        Debug.Log("Lifecycle " + RuntimeHelpers.GetHashCode(this) + " " + text);
    }
}
```

kring:
What about Scene save/load? Saving doesn't raise any of the events I'm looking at. Loading a scene containing this MonoBehaviour triggers the following sequence:

* Object constructor
* OnValidate
* Awake
* OnEnable
* Start
* Update (repeatedly)

And only the field marked `[SerializeField]` survives the save/load process.

kring:
`MonoBehaviour.enabled` returns true before `OnEnable` is called and also after `OnDisable` is called. So I guess it's more like "would like to be enabled" rather than "is actually enabled". 🤷

`MonoBehaviour.isActiveAndEnabled` also returns true before `OnEnable`. But it returns false by the time `OnDisable` is called.

kring:
@j9liu as discussed offline, I had wanted to do some slightly broader reworking on CesiumGeoreference and CesiumGlobeAnchor. But it isn't going all that well, in particular when it comes to preserving backward compatibility with existing scenes (see #188). So rather than continue to wait on that, I'm going to merge this, which is a clear improvement over what's in main and looks very reasonable in general. Thanks!

PR #187, opened 2023-01-25
Add sky controller prefab for sun positioning

### LINKED ISSUE #173: Sun, sky, and atmosphere implementation
Cesium for Unreal includes a CesiumSunSky actor that creates a convincing sky, with some adjustable parameters to aid with geospatial accuracy and lighting customization. We should discuss implementing a similar actor in Cesium for Unity.

### Additional context on CesiumSunSky:
- Much of Unreal's CesiumSunSky came from the Unreal Engine Sun Position Calculator plugin, so there was a smaller initial barrier to creating the actor. Our implementation was a modification of the Sun Position Calculator's SunSky actor.
- The primary modifications to the SunSky actor were hooking it up to the Georeference so that the calculated sun angle changed with scene location. We also made some adjustments to the atmosphere settings to avoid visual artifacts.

### Desired result
To implement a comparable feature in Unity, we would likely need:

- A prefab GameObject to control sky settings. This could be drag-and-dropped into a scene, and even added to the Quick Add panel.
   - A sky controller script that adjusts lighting and shading as needed. It will need to be able to calculate a sun angle based on a given longitude and latitude, date, and time.
   - Available settings should include:
      - Time of day (float, 0-24)
      - Date (Day, month, year)
      - Longitude and Latitude (Set by georeference)
      - Consideration for Daylight Savings Time (This may be optional.)
   - The sky controller should contain a Directional Light gameobject, or be able to reference one from the scene. 
- A sky shader
   - The color of the sky must change depending on the time of day from the sky controller
   - The sky must fade to black as the camera moves away from earth.

### Breakdown of current sky shader options in Unity
Available sky options differ between Unity's High Definition Render Pipeline (HDRP) and Universal Render Pipeline (URP). The HDRP options in terms of sky shading are generally suitable. On the shading side, our implementation would focus on URP, where the deficits are more obvious. 

- URP
   - Skybox material is set per-scene in the Lighting window (Window>Rendering>Lighting)
   - Several preset skybox shader options exist. The most realistic is the Procedural Skybox shader, which allows for a skybox split horizontally between a sky and a ground color. The sky color changes with the angle of the sun, and appears black when the sun is set. 
   - The skybox is very obvious when viewing the earth from space - the background is the skybox's ground color, rather than black. 
image
   - Sun angle for skybox calculations is determined by angle of the scene's primary Directional Light actor.
   - URP is optimized for many devices, including mobile - therefore a physically-accurate atmosphere calculation may be overkill. We can likely achieve suitable atmosphere visual effects through a custom sky shader.
   - A custom sky shader will be needed to handle the transition between earth and space.
- HDRP
   - Sky material is set by
   - A suitable sky and atmosphere effect can be achieved with a Sky and Fog volume placed in the level. The Sky type should be set to Physically Based Sky. 
image 
This means we should not need a custom shader for HDRP.
   - The atmosphere suffers from spotty artifacts when viewing from space. As in Unreal, we may be able to resolve this by adjusting atmosphere settings based on view height. 
   image

### Further considerations

- For URP, because the skybox material is determined by a scene setting rather than an object in the scene, we might need functionality to assign the skybox material through the sky controller gameobject.
- We likely would not want to name the sky implementation "CesiumSunSky", as the SunSky name came from Unreal plugin content. Perhaps "CesiumSkyController" instead. 
- Some research will need to be done to determine the best way to support both URP and HDRP

argallegos:
This implements a sun, sky, and atmosphere solution for Cesium for Unity per https://github.com/CesiumGS/cesium-unity/issues/173. 
Includes the following components:

- A sky controller script that exposes time of day and other settings to the user to control direction of the sun
- A prefab that users can add to their levels to quickly set the required parameters
- A skybox shader for URP that creates an atmosphere blending effect when the camera moves from earth to space. 

It's currently a work in progress. Remaining items:

- [ ] Ensure HDRP compatibility - this is primarily on the shader side, but further work needs to be done to tie the prefab to the HDRP atmosphere options.
- [ ] Implement sun positioning math in cesium-native
- [ ] Sky shader
   - [ ] Improved atmospheric blending
   - [ ] A ground color that reduces the visual impact of missing tiles


kring:
It's not available in any pre-built binaries. If you want to use this, the only option is to build from this PR's branch yourself. It's based on a very old version of Cesium for Unity, though, so you'll probably want to merge in main. In short, it's likely to be a lot of work. Sorry we don't have a simple solution right now.

flankechen:
> It's not available in any pre-built binaries. If you want to use this, the only option is to build from this PR's branch yourself. It's based on a very old version of Cesium for Unity, though, so you'll probably want to merge in main. In short, it's likely to be a lot of work. Sorry we don't have a simple solution right now.

I did try to build cesium from this PR branch and try to run in some cesium example branches, sadly not working. yeap, this looks some old codes. I might not have time for a more deep dive into this. Anyway, looking forward for this to develped and merge into main branch and get release. UE sky looks cool, looking forward for Unity.

asafMasa:
Hi, 
@argallegos are you still working on this PR?
It sounds real nice to have this in Unity

kring:
@asafMasa Alex isn't working on this anymore unfortunately, but we'd be very happy to see someone from the community pick it up and run with it!

baruchInsert-tech:
Hi @kring, I'd like to pick this up and get sun/sky/atmosphere (#173) across the line.

I saw this PR has been sitting since early 2023 and is based on a pretty old version of Cesium for Unity, and that you mentioned you'd be happy for someone from the community to run with it. I've spent some time digging into the problem and I'd like to take it on.

Rather than try to revive this branch as-is, I'm planning to rebase the good ideas here onto current main and rework it into a few smaller, reviewable pieces. I went through the existing code here, the CesiumSunSky actor in cesium-unreal, and the CesiumJS Clock example to make sure I understand what "correct" looks like before writing anything.

Here's the plan I'm proposing:

Phase 1 - Sun + URP sky
- A CesiumSkyController component with date, time of day, and time zone. Latitude/longitude come from the CesiumGeoreference, same as Unreal.
- Sun position from the NOAA solar algorithm (the same one Unreal's Sun Position Calculator uses).
- Orient a directional light from it, drive ambient, and a custom URP skybox shader that fades to black as you leave the surface - basically what this PR already started.

Phase 2 - Clock + UI
- A small CesiumClock (start/current/stop time, multiplier, play/pause, loop) mirroring CesiumJS's Clock, plus a simple on-screen play/pause/speed/scrub UI so the day/night cycle can actually animate.

Phase 3 - HDRP
- Wire the same sun up to HDRP's Physically Based Sky and port Unreal's trick of adjusting the atmosphere ground radius by camera height, which is what fixes the "artifacts from space" problem #173 mentions.

Phase 4 - Integration + polish
- Prefab + Quick Add entry so it's drag-and-drop, tests for the solar math, and visual polish.

A couple of things I ran into that I'd like your input on before I start:

1. This PR's checklist has "Implement sun positioning math in cesium-native". In my prototyping I found it's clean to do entirely in C# (the sun is far enough away that its ECEF direction is effectively the same everywhere on Earth, so no observer position is needed, and it matches how CesiumJS produces sunDirectionWC). Is there a reason you'd prefer it live in cesium-native, or is a C# implementation acceptable?

2. One correctness note: I'd compute the sun direction in ECEF and convert with the georeference's transforms rather than building it in a local East/Up/North frame. The ENU shortcut is only correct in CartographicOrigin placement mode - it silently breaks in TrueOrigin mode - so going through ECEF keeps it correct in both.

If this direction sounds good I'll start on Phase 1 and open it as its own PR so it's easy to review in isolation. Thanks!

kring:
Hi @baruchInsert-tech, thanks for picking this up! Your plan sounds reasonable to me, though @j9liu and @azrogers may want to weigh in as well.

Regarding cesium-native versus C#, the main advantage to putting it in cesium-native is that it allows it to be used across engines. For example, if the sun position were implemented in C# and we wanted to add it to [3D Tiles for Godot](https://cesium.com/blog/2025/05/01/introducing-3d-tiles-for-godot-by-battle-road/), we would need to reimplement it. Adding it to cesium-native makes it more widely reusable.

So I would say that's not a strict requirement, but as long as the burden isn't too unreasonably high for you, I would definitely recommend that you do it.

PR #194, opened 2023-01-30
Access metadata properties based on name

### LINKED ISSUE #165: Access metadata property by name
The only function currently available in CesiumMetadata.cs is `GetProperties()`, which returns an array of type `MetadataProperty`. As a result, the only way to access individual metadata properties is with a `for` loop over the entire array, or by accessing a specific index in the array. So if someone wanted to get only the height property of a selected primitive, they have to get all the metadata, and loop through each element in the array to check whether or not its propertyName = "height".

It would be beneficial to add a function to get a specific metadata property by Property Name. That way, users could instead do something like `GetPropertyByName("height")` instead of getting all properties, than looping.

Potentially also worth discussing: Renaming `GetProperties()` to `GetAllProperties()` for clarity?

joseph-kaile:
fixes #165

the cesium samples script that uses these changes: https://github.com/CesiumGS/cesium-unity-samples/pull/42


Runtime/CesiumMetadata.cs
```diff
@@ -9,29 +10,60 @@ namespace CesiumForUnity
...
         /// The information to pass to this function can be obtained using the <code>Physics.Raycast</code>
         /// function.
         /// </remarks>
-        public MetadataProperty[] GetProperties(Transform transform, int triangleIndex)
-        {
-            int numberOfProperties = getNumberOfProperties(transform);
-            MetadataProperty[] properties = new MetadataProperty[numberOfProperties];
-            for(int i = 0; i < numberOfProperties; i++){
-                properties[i] = new MetadataProperty();
+        public Feature[] GetFeatures(Transform transform, int triangleIndex){
```
kring, line 43: Did you consider implementing this entire function in C++? I guess Reinterop might struggle with the Dictionary?

kring (commented):
Looks great, @joseph-kaile! My only concern, other than the question below, is that `MetadataProperty` is backed by a C++ instance. This makes it a fairly heavyweight object that (ideally) would be explicitly disposed so it doesn't need to be finalized. It's ok if it ends up being a separate PR or even a future change, but I think it's worth considering whether we can improve this. For example, all `MetadataProperty` instances could instead each have a reference to a single instance of a C++-backed, along with whatever context information (name, index, etc.) is needed to find the property itself. Then _static_ partial methods can do the name, value, etc. lookup by delegating to the single instance.

Runtime/CesiumMetadata.cs
joseph-kaile: I'm getting stuck at how to create the new objects, new MetadataProperty[] and new Feature[] can only be called in C#.

joseph-kaile:
> ed to be finalized. It's ok if it ends up being a separate PR or even a future change, but I think it's worth considering whether we can

Yes, I see how I can do this and it can be part of this PR.

kring:
Thanks @joseph-kaile! I think this could still be more efficient, not requiring a C++ _instance_ to back `CesiumFeature`. But I think we can fix that in the future without a breaking change, and this is clearly better than what we had before. So I'm going to merge it. Thanks for reworking it!

PR #197, opened 2023-02-02
Add support for point clouds in Unity

j9liu:
This PR builds off of https://github.com/CesiumGS/cesium-native/pull/591 and adds support for rendering point clouds (`pnts` content) in Unity.

Some screenshots:
| Melbourne Point Cloud |
| ----------------------- |
| image |

| Mount St. Helen | Church |
| ------------------ | -------- |
| image | image |

[CesiumJS sandcastle with the above models for reference](https://sandcastle.cesium.com/?src=3D%20Tiles%20Point%20Cloud%20Shading.html)

### DIFF
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -1,5 +1,11 @@
 # Change Log
 
+### v0.3.0
+
+##### Additions :tada:
+
+- Added support for rendering point clouds (`pnts`).
+
 ### v0.2.0
 
 ##### Breaking Changes :mega:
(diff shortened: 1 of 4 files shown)


lilleyse:
@j9liu this is exciting to see!

The colors look a bit faded and I wonder if you need to do an sRGB -> linear conversion in cesium-native. `pnts` is assumed to be sRGB whereas glTF vertex colors are assumed to be linear. I'm not sure what Unity expects as its input.

Peter ran into a similar issue in CesiumJS: https://github.com/CesiumGS/cesium/pull/9978#issuecomment-1005164892

j9liu:
@lilleyse Thanks for the catch! It does look more accurate with the colors converted to linear space:
image

Will push my changes to `cesium-native` once I update the unit tests as well.

kring:
Looks great!

Issue #207, opened 2023-02-14
IonTokenTroubleshootingWindow crashes


kring:
On any tileset, click the "Troubleshoot Token" button on the Inspector panel. The troubleshooting panel appears, but then Unity crashes with this strack trace:

```
Obtained 31 stack frames
0x00007ffaa27bdfa8 (CesiumForUnityNative-Editor) [traits.h:109] async::detail::invoke_fake_void<<lambda_1eaefc7302ab98935558272345ed73d0>,CesiumIonClient::Response<CesiumIonClient::Profile> > 
0x00007ffaa27bae0b (CesiumForUnityNative-Editor) [task_base.h:398] async::detail::task_func<CesiumAsync::CesiumImpl::ImmediateScheduler<CesiumAsync::CesiumImpl::QueuedScheduler>,async::detail::continuation_exec_func<CesiumAsync::CesiumImpl::ImmediateScheduler<CesiumAsync::CesiumImpl::QueuedScheduler>,async::task<CesiumIonClient::Response<CesiumIonClient::Profile> >,async::detail::fake_void,<lambda_1eaefc7302ab98935558272345ed73d0>,std::integral_constant<bool,1>,0>,async::detail::fake_void>::run 
0x00007ffaa27a04be (CesiumForUnityNative-Editor) [scheduler.h:165] async::detail::schedule_task<CesiumAsync::CesiumImpl::ImmediateScheduler<CesiumAsync::CesiumImpl::QueuedScheduler> > 
0x00007ffaa27a03c8 (CesiumForUnityNative-Editor) [task_base.h:101] async::detail::task_base::run_continuation<CesiumAsync::CesiumImpl::ImmediateScheduler<CesiumAsync::CesiumImpl::QueuedScheduler> > 
0x00007ffaa27baf8c (CesiumForUnityNative-Editor) [task_base.h:417] async::detail::task_func<CesiumAsync::CesiumImpl::ImmediateScheduler<CesiumAsync::CesiumImpl::QueuedScheduler>,async::detail::continuation_exec_func<CesiumAsync::CesiumImpl::ImmediateScheduler<CesiumAsync::CesiumImpl::QueuedScheduler>,async::task<CesiumIonClient::Response<CesiumIonClient::Profile> >,async::detail::fake_void,<lambda_1eaefc7302ab98935558272345ed73d0>,std::integral_constant<bool,1>,0>,async::detail::fake_void>::schedule 
[...]

PR #209, opened 2023-02-15
Add `staticOnly` flag to `ReinteropNativeImplementation` attribute

kring:
Before this PR, static partial methods would be delegated to a static method on a C++ class, and non-static partial methods would be delegated to an instance method on the C++ class. If there were any non-static methods at all, then a C++ instance would be created whenever a C# instance was, and the C# class would have a finalizer to ensure the C++ instance was eventually destroyed in order to avoid a memory leak.

This kind of made sense, but was inefficient in some cases. Sometimes, such as for `CesiumGlobeAnchor`, we don't want to have a C++ instance, because it would be costly to have a finalizer, and our C++ implementation doesn't have any state separate from the C# class anyway. But it's still useful to have partial instance methods so that the C++ code can access the state on the C# instance. This led us to kind of ugly workarounds, like declaring the partial method static but then passing the C# instance as the first parameter. And it would be all too easy for a future developer (probably future me) to add a non-static partial method to `CesiumGlobeAnchor` and suddenly make it finalizable, without realizing the cost.

This PR fixes both of those problems by adding a new parameter, `staticOnly` to the `[ReinteropNativeImplementation]` attribute. When set to true, Reinterop will _never_ create a C++ instance to back the C# one. Non-static partial methods are allowed, but will be implemented in C++ by a static method. The static method will receive the C# instance as its first parameter, just like all partial methods.

This way, we can write our partial methods in a more natural way, and there's no danger of accidentally making a class like `CesiumGlobeAnchor` finalizable without explicitly changing `staticOnly` to false.

I've applied this new parameter to `IonTokenTroubleshootingWindow`, `CesiumWgs84Ellipsoid`, and `CesiumGlobeAnchor`.

### DIFF
--- a/Reinterop~/MethodsImplementedInCpp.cs
+++ b/Reinterop~/MethodsImplementedInCpp.cs
@@ -19,7 +19,7 @@ namespace Reinterop
             CppType objectHandleType = CppObjectHandle.GetCppType(context);
 
             // We only need a C++ instance if any partial methods are non-static
-            bool needsInstance = item.MethodsImplementedInCpp.Any(m => !m.IsStatic);
+            bool needsInstance = item.ImplementationStaticOnly != false && item.MethodsImplementedInCpp.Any(m => !m.IsStatic);
             result.CSharpPartialMethodDefinitions.needsInstance = needsInstance;
 
             if (needsInstance)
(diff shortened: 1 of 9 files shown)


Reinterop~/MethodsImplementedInCpp.cs
```diff
@@ -19,7 +19,7 @@ public static void Generate(CppGenerationContext context, TypeToGenerate item, G
             CppType objectHandleType = CppObjectHandle.GetCppType(context);
 
             // We only need a C++ instance if any partial methods are non-static
-            bool needsInstance = item.MethodsImplementedInCpp.Any(m => !m.IsStatic);
+            bool needsInstance = !item.ImplementationStaticOnly && item.MethodsImplementedInCpp.Any(m => !m.IsStatic);
```
j9liu, line 22: Is it possible to for someone to (incorrectly) declare a static-only implementation and declare a non-static partial method? Is there a check for that ?

j9liu (commented):
@kring looks good! I just had one question. I noticed a comment that has become irrelevant but I'll push a commit to remove it.

Reinterop~/MethodsImplementedInCpp.cs
kring: That's actually allowed! It just means the the _static_ C++ implementation will receive the C# _instance_ as its first parameter.

j9liu: ah, I think I understand! easy to lose track of what's static and what's an instance 😅

j9liu:
Also for the record, I couldn't reproduce #207 on this branch. Maybe it'll come up in one of the others 🤔

Issue #210, opened 2023-02-15
Reinterop: Allow C++ implementation classes to access private fields / properties / methods of C# class


kring:
When using the `[ReinteropNativeImplementation]` attribute, the C++ and C# classes are conceptually one class, just with parts implemented in two different languages. To make that more seamless, it would be nice if the C++ code could access private members of the C# class. Currently the only way to do this is by declaring the private members "internal" instead.

PR #211, opened 2023-02-15
Reinterop: Fix code gen for non-static field access.

kring:
This is a PR into #210, so merge that first.

Previously the generated C++ code for accessing non-static fields would not compile.

### DIFF
--- a/Reinterop~/Fields.cs
+++ b/Reinterop~/Fields.cs
@@ -101,9 +101,9 @@ namespace Reinterop
             if (!field.IsStatic)
             {
                 interopGetParameters = $"void* thiz";
-                interopGetParametersCall = $"{definition.Type.AsParameterType().GetConversionToInteropType(context, "thiz")}";
+                interopGetParametersCall = $"{definition.Type.AsParameterType().GetConversionToInteropType(context, "(*this)")}";
                 interopSetParameters = $"void* thiz, " + interopSetParameters;
-                interopSetParametersCall = $"{definition.Type.AsParameterType().GetConversionToInteropType(context, "thiz")}, {interopSetParametersCall}";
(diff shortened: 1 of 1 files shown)


j9liu:
I'll take a closer look at the code tomorrow, but I tested out different types of fields (public, internal, static, blittable struct) and it looks like it works well 🎉

Reinterop~/Fields.cs
```diff
@@ -156,12 +163,38 @@ private static void GenerateSingleFieldAccessors(CppGenerationContext context, T
...
             ));
 
+            string[] invocation = new[]
+            {
+                $"auto result = Field_get_{field.Name}({string.Join(", ", interopGetParametersCall)});",
+                $"return {getType.GetConversionFromInteropType(context, "result")};"
+            };
+            if (hasStructRewrite)
+            {
+                if (fieldType.Kind == InteropTypeKind.Nullable)
```
j9liu, line 173: When I do a nullable non-struct field in a class (for example, I added `public int? test2` to `CesiumGeoreference`, it's not generating code with `std::optional`, it's doing this instead:
```
  public: ::DotNet::System::Nullable1<::std::int32_t> test2() const;
  public: void test2(const ::DotNet::System::Nullable1<::std::int32_t>& value) const;
```

Is that intentional?

j9liu (commented):
@kring looks good, I just noticed one potential bug. Otherwise, everything checks out.

Reinterop~/Fields.cs
kring: Not intentional, but I think it can be addressed in a separate PR. I wrote an issue for it:
https://github.com/CesiumGS/cesium-unity/issues/225

PR #212, opened 2023-02-15
Generate Reinterop cpp files into namespace subdirectories.

kring:
This is a PR into #211 so merge that first.

Otherwise, `UnityEngine.Quaternion` and `Unity.Mathematics.quaterion` both get generated to the same source file (`quaternion.cpp`) on a case-insensitive file system.


j9liu:
Looks good, thanks @kring !

PR #214, opened 2023-02-15
copy response headers to enable caching

### LINKED ISSUE #208: Caching Asset Accessor not caching entries
There is no "Expires" in the response header and so entries are not cached.

joseph-kaile:
Fixes #208

### DIFF
--- a/native~/Shared/src/UnityAssetAccessor.cpp
+++ b/native~/Shared/src/UnityAssetAccessor.cpp
@@ -8,6 +8,9 @@
 #include <DotNet/CesiumForUnity/Helpers.h>
 #include <DotNet/CesiumForUnity/NativeDownloadHandler.h>
 #include <DotNet/System/Action1.h>
+#include <DotNet/System/Collections/Generic/Dictionary2.h>
+#include <DotNet/System/Collections/Generic/Enumerator0.h>
+#include <DotNet/System/Collections/Generic/KeyValuePair2.h>
 #include <DotNet/System/Environment.h>
 #include <DotNet/System/OperatingSystem.h>
 #include <DotNet/System/String.h>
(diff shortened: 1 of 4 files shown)


native~/Shared/src/UnityAssetAccessor.cpp
```diff
@@ -39,13 +42,16 @@ class UnityAssetResponse : public IAssetResponse {
...
-        request.GetResponseHeader(System::String("Content-Type"));
-    if (contentTypeHeader != nullptr) {
-      this->_contentType = contentTypeHeader.ToStlString();
-      this->_headers.emplace("Content-Type", this->_contentType);
+    System::Collections::Generic::Enumerator0 enumerator =
+        request.GetResponseHeaders().GetEnumerator();
+    while (enumerator.MoveNext()) {
+      std::string key = enumerator.Current().Key().ToStlString();
+      std::string value = enumerator.Current().Value().ToStlString();
+      if (value == "Content-Type") {
```
kring, line 50: Should this be `key == "Content-Type"`? Also HTTP headers are case-insensitive, so unless Unity is normalizing them we need to do a case-insensitive comparison. (The `GetResponseHeader` method we used before handles this automatically - maybe that easiest thing is to go back to retrieve the Content-Type)

native~/Shared/src/UnityAssetAccessor.cpp
```diff
@@ -39,13 +42,16 @@ class UnityAssetResponse : public IAssetResponse {
...
+        this->_contentType = value;
+      }
+      this->_headers.emplace(key, value);
```
kring, line 53: I think using std::move with both key and value here will be beneficial (avoid two heap allocations).

kring (commented):
Excellent, good catch Joseph! A couple of suggestions below.

kring:
Thanks @joseph-kaile!

PR #215, opened 2023-02-15
Organize material graphs

argallegos:
Small adjustments to improve the appearance of default tileset materials. 

The CesiumDefaultTilesetShader and CesiumUnlitTilesetShader have been recategorized so they are now in the Cesium category rather than the Shader Graph categories. This is noted in the changelog.

Node graphs have been cleaned up and grouped. This way, sections of the shader can be dragged around as a group and it should be easier for users who want to build off these materials.
image
This is a purely visual change and listing it in the changelog seemed a bit pedantic. As far as I'm aware, it shouldn't cause any breaking changes.

### DIFF
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -7,6 +7,10 @@
 - Added support for rendering point clouds (`pnts`).
 - Metadata features are now separated based on feature tables. Properties can now be accessed by name.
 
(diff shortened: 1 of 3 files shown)


kring:
Thanks @argallegos!

Issue #180, opened 2023-01-17
Changing material parameters does not update material for all tiles


argallegos:
When changing the parameters of a 3D Tileset's material, tiles won't update to reflect these changes uniformly. 

Here's a visualization - a single simple material was assigned to this tileset, and the color of the material was changed three times. Only some tiles reflect the changes. It seems like only new tiles will change, while tiles that are already loaded do not update. 

![image](https://user-images.githubusercontent.com/39537389/213016538-8e1653d7-e1c5-453b-a4b6-0b82fa7ce94e.png)

Note that using the "Refresh Tileset" button does not resolve this. 

Steps to reproduce:
1. Create a new material using the Universal Render Pipeline/Lit shader.
2. Assign the new material to a 3D Tileset, like Cesium World Terrain. 
3. Select the material and change the Base Color to a different color. 
4. Move the camera.

From https://community.cesium.com/t/how-to-create-a-shader-containing-the-original-texture-of-the-cesium-world-terrain/21871/2

argallegos:
Side note - At first glance, I'd assume this is happening because each tile uses its own instance of the material, and I don't think material instances phone home to the original material in the same way that Unreal materials do. Possibly, the best solution here would be to use [MaterialPropertyBlocks](https://docs.unity3d.com/ScriptReference/MaterialPropertyBlock.html) for all 3D Tileset materials in order to update all instances at once.

PR #218, opened 2023-02-20
Add support for point cloud attenuation

j9liu:
This PR adds support for attenuating point clouds in Cesium for Unity. The `CesiumPointCloudShading` class is serialized with the `Cesium3DTileset` and acts as a counterpart to CesiumJS's [`PointCloudShading`](https://github.com/CesiumGS/cesium/blob/9295450e64c3077d96ad579012068ea05f97842c/packages/engine/Source/Scene/PointCloudShading.js) class.

Settings for below: Max SSE = 16, Geometric Error Scale = 0.5, Maximum Attenuation = 4

| With attenuation | Without attenuation |
| ------------------- | --------------------- |
| image | image |
| image | image |

Settings for below: Max SSE = 16, Geometric Error Scale = 0.5, Maximum Attenuation = 4, Base Resolution = 0.05 (same as in [CesiumJS Sandcastle](https://sandcastle.cesium.com/?src=3D%20Tiles%20Point%20Cloud%20Shading.html)

| With attenuation | Without attenuation |
| ------------------- | --------------------- |
| image | image |
| image | image |
| image | image |

It's possible to change the values via the `Cesium3DTileset` editor without refreshing the entire tileset.

PointCloudShadingParameters

**TODO**:
- [ ] ~Check out potential bug with shadow mapping (shadows appear much bigger than the points because the shadow map used is a smaller size)~
- [x] Create lit material + test on point cloud with normals
- [ ] ~Create shader for HDRP (hopefully mostly copy-paste)~
- [ ] Check compatibility with other devices
  - [x] MacOS / iPhone
  - [ ] Android
  - [x] VR
- [x] Optimization / cleanup
  - [x] Performance analysis
  - [ ] ~Consider replacing the original mesh renderer / filter entirely with `CesiumPointsRenderer` to avoid duplication of data~


j9liu:
So... my previous profiling is actually invalid, because the Unity profiler only displays the last 300 frames of data and I incorrectly assumed that when you stopped it, it would display the entire history. So I don't actually have good numbers on the performance before I refactored it. However, I made some changes that I think are better in the long run and can be intuitively justified.

First, instead of copying data in `ComputeBuffer`s only when we need it, the `GraphicsBuffer` containing all of the vertex data is directly used. The complexity with this is that we'll have to include `#defines` in the shader for handling the different combination of attributes.

Second, I learned that if you want to draw multiple objects with the same material, just different property values, you can use one instance of the material, then have each object use a different [`MaterialPropertyBlock`](https://docs.unity3d.com/ScriptReference/MaterialPropertyBlock.html). This made me gravitate towards a model where `CesiumPointCloudShading`, in addition to the public API, also has an internal component where it manages one material per point cloud tileset. Then, all of the `CesiumPointCloudRenderer`s just have to reference that material without instantiating their own. I think there is at least _some_? performance improvement because I'm not seeing a bunch of `Material.SetPass` calls anymore in the profiler.

However, I think the main culprit is still that `Graphics.DrawProcedural` is submitted one draw call per tile. Unity does batching under-the-hood, so when we're drawing the points with the regular `MeshRenderer`, we're actually only submitting a couple of draw calls. The regular rendering time isn't actually increased _that_ much with attenuation on, but the main culprit of the framerate decrease is the shadows pass:

image

In this profile, the higher GPU usage corresponds to attenuation on, and lower GPU usage corresponds to attenuation off. The red graph corresponds to the "Other" category, which I guess for Unity is the normal way of lighting and rendering points. The orange corresponds to the "Shadows/Depth" category. With attenuation on; shadows take around 7.6 ms and opaque object rendering takes 8.5 ms. This is pretty terrible compared to having no attenuation, which takes about 2.5ms to render shadows and 4.875 ms to render opaque objects.

So I guess the best way to improve performance is to find a way to batch draw calls together, but I'm not sure how to best do this. In my implementation it seems like `CesiumPointCloudShading` would have to manage all the `CesiumPointCloudRenderer`s of the object and set / remove data from a compute buffer, then draw it in one `Graphics.DrawProcedural` call. It _will_ involve duplication of data, so there's that memory vs. runtime performance tradeoff, and that's assuming that the runtime performance even improves with the compute buffer size potentially changing. But I'll try it before I knock it.

j9liu:
> So I guess the best way to improve performance is to find a way to batch draw calls together, but I'm not sure how to best do this. In my implementation it seems like `CesiumPointCloudShading` would have to manage all the `CesiumPointCloudRenderer`s of the object and set / remove data from a compute buffer, then draw it in one `Graphics.DrawProcedural` call. It _will_ involve duplication of data, so there's that memory vs. runtime performance tradeoff, and that's assuming that the runtime performance even improves with the compute buffer size potentially changing. But I'll try it before I knock it.

Actually, this is probably unreasonable to try and implement when we don't even know if it'll help performance. If the points didn't have tile-specific parameters, it'd be doable. But the whole point of point attenuation is to use the tile's geometric error to scale the points appropriately. If we batched all of the point cloud in one buffer, we'd also need an extra attribute to distinguish which tile each point belongs to, and then retrieve the parameters from an array of parameters for all tiles. Never mind the fact that we are constantly changing the buffer contents depending on what tiles are visible. So that's really complicated for what is probably little reward.

At this point, maybe it's just worth leaving the `MeshRenderer` attached as-is to the tile, and keeping the `CesiumPointCloudRenderer` as an addendum. The `Material` and `GraphicsBuffer` changes are probably still helpful, but there's still going to be a dramatic draw call overhead that from what I can see, we can't easily avoid.

j9liu:
I retried the `DrawMeshInstancedProcedural` approach, but I still could not get working because `unity_InstanceID` was always zero. Even with all of the other hacks people seemed to have gotten to work. I was considering trying to implement the function from ShaderGraph, since it has an Instance ID node and I figured I may be doing something wrong from manual HLSL code that was preventing the instance ID from being properly defined. But before I went down that rabbit hole, I ran the profiler with my (incorrectly rendering) attenuated points just to see how they performed, and...

image

😬 It's even worse. It looks like the instancing is adding a lot more computation on the GPU, and is more than tripling the amount of time it takes to do both the opaque and the shadow passes. Granted, these stats are with incorrect rendering, but it's still drawing the same number of vertices (quad mesh * point count). So even if I miraculously got the instance ID to work, I don't think that'd have any impact on performance.

I'll also quote Unity's documentation on [GPU instancing](https://docs.unity3d.com/Manual/GPUInstancing.html) in my argument against continuing to try and implement this:
> Meshes that have a low number of vertices can’t be processed efficiently using GPU instancing because the GPU can’t distribute the work in a way that fully uses the GPU’s resources. This processing inefficiency can have a detrimental effect on performance. The threshold at which inefficiencies begin depends on the GPU, but as a general rule, don’t use GPU instancing for meshes that have fewer than 256 vertices.

j9liu:
At @kring's suggestion I tried using `DrawMesh` instead of `DrawProcedural`, which people across Unity forums seem to have reported as reliably slow. But I'm not actually sure there is much improvement by using `DrawMesh` alone:

image

The large spike when I turned attenuation on scales this graph pretty low, but the numbers are similar. Without attenuation, a render pass takes around 4.9 ms to draw the points and 2.4 ms to render shadows. With attenuation on, it still takes around 8ms to draw the points, and 6.4 ms to render shadows. It's technically better than `DrawProcedural`, but not by much, and it also adds 0.35GB extra mesh data to render with attenuation.

image

That said, I wonder if we can just get the `MeshRenderer` to handle drawing the meshes itself, instead of doing a `Graphics.DrawXXX` call. If we can take advantage of Unity's built-in batching system and reduce the GPU time by a couple milliseconds, I think it will be worth it. But I have to see if I can get this idea to work in the first place.

j9liu:
I wanted to standardize my profiler results because I wasn't consistent about the amount of tiles that were showing on the screen for my earlier tests, and I want to make sure the work I'm doing is legitimately improving performance. These stats are with Melbourne Point Cloud, max SSE 16, CWT disabled, no movement on Dynamic Camera. 

In these tests, data was always passed to the shader as one mesh vertex buffer, instead of duplicating the data in separate `ComputeBuffer`s. I also made every single `CesiumPointCloudRenderer` instantiate a material.

| Method | Without Attenuation | With Attenuation | Profiler Image | 
| ---------- | --------------------- | ------------------- | ---------- | 
| No implementation (empty `CesiumPointCloudRenderer` behaviour) | Consistently above 60 FPS. 10.6ms in opaque pass, 2.2 ms in shadow pass. In `DrawOpaqueObjects`, `RenderLoop.ScheduleDraw` has 1 draw call that takes 5ms. Mesh data is 214.1MB. Vertices count is 16.99M, but for some reason triangles count is 5.66M ? | N/A. | image |
| `Graphics.DrawProcedural` (old code pasted in latest commit, with unnecessary things commented out) | Consistently above 60FPS. Similar GPU stats as above. Mesh data is 214.1MB | Below 60 FPS. 10.81ms in opaque pass and 7ms in shadow pass. `RenderLoop.ScheduleDraw` has 69 draw calls that take 7.7 ms total. Mesh data is still 214.1MB | image |
| `Graphics.DrawMesh` (b08d2add14f8263df90c529913ee802892b65077) | Consistently above 60 FPS. Similar GPU stats as above. Mesh data is 0.55GB (duplication of data, since this was with the original meshes as well. so the new expanded point meshes might be 335.9MB on their own) | Below 60 FPS. 11.4ms in opaque pass, 6.2ms in shadow pass. `RenderLoop.ScheduleDraw` has 69 draw calls that take 8.40ms total. Mesh data is 0.89GB (why does it go up here?) | image |

I'm not sure what caused `DrawProcedural` to take 19ms per frame before, but now it seems fine... better, even. `DrawMesh` is not outperforming it in a way that justifies the extra memory.

I still need to try using the `MeshRenderer` itself and seeing if the batching will work on it. If the batching shaves off that extra shadow pass time, it may be worth it. But it's not working well with the clip-space offsets. So if it's not even possible, then... `DrawProcedural` it is.

j9liu:
I was able to get attenuation to show up with `MeshRenderer`. It turns out that the built-in shader variables `_ScreenParams` and `_ProjectionParams` aren't defined / bound in shaders used by `MeshRenderer`. For the sake of testing, I hardcoded the screen-size so I could profile performance. I saw that the draw calls _were_ being batched: under `RenderLoop.ScheduleDraw` the profiler hierarchy includes `RenderLoop.DrawSRPBatcher`. But the `RenderLoop.ScheduleDraw` calls in the opaque pass were still taking twice as long with attenuation.

This leads me to think that something is just inherently slow with the shader itself, or at least how it interacts with the shadow pass. I thought perhaps the vertex ID division / modulo in the shader could be expensive, but I saw no improvement when using the mesh's UV attribute to store the corners.

| Method | Without Attenuation | With Attenuation | Profiler Image | 
| --------- | ----------------------- | ------------------ | ---------------- |
| `Graphics.DrawProcedural` with individual material (redid today for comparison) | Consistently above 60 FPS. 10.6ms in opaque pass, 2.4 in shadow pass. `RenderLoop.ScheduleDraw` has 1 draw call that takes 4.83ms. Mesh memory is 199.7MB. | Below 60 FPS. 11.15 ms in opaque pass, 7.5 in shadow pass. `RenderLoop.ScheduleDraw` has 64 draw calls that take 7.930ms. Mesh memory is still 199.7MB. | image |
| `MeshRenderer` with expanded point mesh and individual material | Consistently above 60FPS. 10ms in opaque pass, 2.2 in shadow pass. `RenderLoop.ScheduleDraw` has 1 draw call that takes 5ms. Mesh memory is at 0.59GB | Below 60 FPS. 12.5ms in opaque pass, 7ms in shadow pass. `RenderLoop.ScheduleDraw` has 1 draw call that takes 10ms. Mesh memory is 0.98 GB | image |

So I've decided I'm going with `Graphics.DrawProcedural`. This offers practically the same performance as drawing with a mesh, without the overhead of generating and storing additional mesh data. I also tested if using one material (as opposed to an individual material for each `CesiumPointCloudRenderer` had any performance impact, and it made a _bit_ of a difference:

| Method | Without Attenuation | With Attenuation | Profiler Image |
| --------- | ----------------------- | ------------------ | ---------------- |
| `Graphics.DrawProcedural` with shared material | Consistently above 60 FPS. 10.2ms in opaque pass and 2.3ms in shadow pass.     `RenderLoop.ScheduleDraw` has 1 draw call that takes 4.726ms. | Below 60 FPS. 10.8ms in the opaque pass and 6.8ms in shadow pass. `RenderLoop.ScheduleDraw` has 64 draw calls that take 7.650ms. | image | 

It's a small improvement... but taking a step back, I don't actually know if you can guarantee `.pnts` files having the same format across an entire tileset (each tile is a different `.pnts`, some `.pnts` could have a different constant color, for example), so I'm reverting this change. It's possible to have `CesiumPointCloudRenderer` host multiple materials for differences in the tiles but this might be better to rework when we address #180.

j9liu:
Was testing out HDRP today. With a couple keyword changes, attenuation works but it sticks out pretty noticeably:

| Without Attenuation | With Attenuation |
| - | - |
| image | image |

It's true that even without attenuation, the fog on the horizon is rendering on top of the buildings. But it looks like the `CesiumUnlitTilesetMaterial` in HDRP is not as "unlit" as the `CesiumUnlitPointCloudMaterial`.

Of course there's little documentation for how to write an HDRP shader, so I don't know what extra work I'd need to do to fix the appearance. Maybe I _have_ to implement the shader in shader graph for attenuation to look good in HDRP.

j9liu:
> Maybe I _have_ to implement the shader in shader graph for attenuation to look good in HDRP.

I need to be better at documenting my process, because I'm just rediscovering the reasons why I didn't use shader graph in the first place 🙃

Shader graph outputs positions in object space, clip space, and there's no way to calculate the inverse of the projection / view matrices in shader graph. (It'd also be more expensive because we're basically duplicating the number of matrix multiplications we need for the same result.) I also revisited some tutorials where people were able to use compute buffers in shader graph, but they were actually accomplishing it through procedural instancing (see [here](https://catlikecoding.com/unity/tutorials/basics/compute-shaders/) and [here](https://bootcamp.uxdesign.cc/point-cloud-rendering-with-unity-1a07345eb27a)). So to implement attenuation in shader graph, we'd need to procedurally instance a quad with different ways of computing the quad sizes in object space. It's _probably_ doable but I don't know if it's worth the conversion at this point. I'm also not assured of the efficiency of procedural instancing given my earlier profiling results with `DrawMeshInstancedProcedural`.

So I think I'll just try and learn how to write an HDRP shader to make attenuation look better in it. Which there doesn't seem to be a lot of documentation for, but I'll try to look through the shaders in the [ScriptableRenderPipeline repo](https://github.com/UnityTechnologies/ScriptableRenderPipeline/tree/master/com.unity.render-pipelines.high-definition/HDRP/Material) to get an idea.

On the bright side, got attenuation to work with normals :')
PointCloudNormals

j9liu:
> Shader graph outputs positions in object space, clip space, and there's no way to calculate the inverse of the projection / view matrices in shader graph.

I was wrong; there's a Transformation Matrix node that contains inverse view / projection / view-projection matrices. (They're conveniently not listed under the [built-in shader variables](https://docs.unity3d.com/Manual/SL-UnityShaderVariables.html) or in the shader variables file, so I didn't think they existed.) So I did reattempt to get a shader graph working, with the information from the `StructuredBuffer` being retrieved from a custom function node. But I couldn't figure it out. 

For the record, I tried to debug by hardcoding a fixed position (0, 0, 0), hoping to draw the quad with the offsets in object space. This really slowed down the editor when you look at the origin (and it even crashed Unity a few times because the GPU was receiving too much work), but nothing was rendering. Multiplying the positions with my given `_worldTransform` variable didn't help either. The `_inVertices` buffer _did_ exist in the shader graph material; maybe the vertex ID node wasn't working?

My attempts to copy and rewrite the `Lit` shader from HDRP were also in vain, even with a hardcoded color I just couldn't get anything to draw. I tried to copy the `LightLoop` function into the working (but incorrectly lit) HDRP shader I got, but the `#include` dependencies are terrible to manage. It seems like you can never use `UnityCG.inc` inside of an HDRP project, because everything that an HDRP function depends on involves a `Common.hlsl` file that contains duplicate defines as `UnityCG.inc`. It's a big hassle to figure out, not to mention that it's a fragile implementation because the shader code could change from version to version.

So I'm shelfing the HDRP implementation for now; I don't want to hold up this feature any longer. We can release it for URP first and gauge interest in an HDRP version.

joseph-kaile:
@j9liu , I'm seeing some strange behavior in mac. I replaced the Melbourne tileset in the sample scene with the point cloud version. I do see the points, however when I move the camera around, random colors jump around in the screen.
out
It feels like its a precision issue.

j9liu:
@joseph-kaile That's really strange, because it looks like Attenuation is off in the Cesium inspector. Do you mind checking
1. what a normal point cloud looks like
2. what happens if you toggle any of the settings under "Point Cloud Shading" in the tileset inspector?

joseph-kaile:
When I enabled point cloud attennuation, most of the colors go away. With cesium globe terrain enabled as well, it looks like there is Z-fighting.

When spinning the camera in place, there are no visual artifacts. When I move the camera, the screen flashes with large distorted geometry. I think this is related to globe shifting.

j9liu:
@joseph-kaile and I debugged offline and found that this only happens with basic rendering of point clouds. See #241 for more details.

The attenuation itself works well, and because of the artifacts, it's the only way to really see the point cloud on Macs at the moment 😅  thanks for testing Joseph.

j9liu:
@joseph-kaile helped me fix attenuation in VR, since it was only showing up in one eye. The last few commits address this, so now it works for both multi-pass and single-pass instanced VR rendering. Thanks Joseph!

kring:
My 🥔 GPU is an AMD Radeon RX550. It's an AMD CPU also (Ryzen 9 3900X), FWIW. If you're definitely not seeing what I saw above, I'll do a bit of debugging.

j9liu:
@kring yeah, I don't think either @joseph-kaile or I see what you're seeing. I played around with the screen space error on that tileset but it didn't make a difference, I could still see the expected distribution of points at each value.

joseph-kaile:
Yes, I have an NVIDIA card and I don't see those issues.

image
image
image
This is after playing with the settings:
image

kring:
If I switch Unity to DX12 (which is apparently experimental in 2021.3) instead of DX11, it renders correctly! I'm keep digging to see if I can figure out what's going wrong in DX11.

One other thing I noticed though is that the attentuation seems different in Unity versus CesiumJS. For example, if I use Joseph's settings in his last screenshot above in CesiumJS, the points are really big!

image


But in Joseph's screenshot and also on my own system in DX12 where it works, the points are much smaller.

Is it expected that the attentuation should work this differently in the two products?

kring:
Ok, I think I figured it out. The point rendering code a small, subtle thing that is technically illegal in DirectX 11, but NVidia apparently lets it slide.  We have code like this:

```cs
this._mesh.vertexBufferTarget |= GraphicsBuffer.Target.Structured;
this._meshVertexBuffer = this._mesh.GetVertexBuffer(0);
```

And before this is executed the `vertexBufferTarget` is `GraphicsBuffer.Target.Vertex`, because this is a mesh.

But in [Unity's docs for this Target enum](https://docs.unity3d.com/ScriptReference/GraphicsBuffer.Target.html), it says:

> DirectX 11 does not allow [Index](https://docs.unity3d.com/ScriptReference/GraphicsBuffer.Target.Index.html) or [Vertex](https://docs.unity3d.com/ScriptReference/GraphicsBuffer.Target.Vertex.html) buffers to also be [Structured](https://docs.unity3d.com/ScriptReference/GraphicsBuffer.Target.Structured.html). For compute shader mesh data access with DirectX 11 compatibility, use [Raw](https://docs.unity3d.com/ScriptReference/GraphicsBuffer.Target.Raw.html).

Oops! The result - at least on AMD GPUs - is some weird stride problem where `_inPoints[pointIndex]` returns the wrong thing. Here's RenderDoc showing the vertexIndex, pointIndex, and vertex position from _inPoints that we see in the vertex shader:

image

The first position is fine. The next one, instead of starting at the next vertex, seems to read the Y position into X, the Z position into Y, and NaN into Z. It's like the stride is 4 instead of 16. In any case, these bad positions cause us to compute NaNs and the points don't appear. Every fourth point is fine, so we can still see about a quarter of the points.

I tried to fix this by removing the `Vertex` target. But Unity seems to ignore attempts to remove this target from a mesh buffer (nevermind that would probably break the non-attentuated path). So then I tried copying the data to a new buffer with the correct target:

```cs
GraphicsBuffer sourceBuffer = this._mesh.GetVertexBuffer(0);
this._meshVertexBuffer = new GraphicsBuffer(GraphicsBuffer.Target.Structured | GraphicsBuffer.Target.CopyDestination, sourceBuffer.count, sourceBuffer.stride);
Graphics.CopyBuffer(sourceBuffer, this._meshVertexBuffer);
```

And it works! All the points appear as expected.

I'm not sure what the ultimate solution is here. At the very least we need to actually free that buffer when the tile is destroyed. And maybe there's a way to avoid the copy? Or maybe not.

j9liu:
Thanks for finding the issue @kring! I want to see if we can avoid copying the buffer, so I pushed a fix that tries to use `Buffer` in DX11 instead -- can you see if that works?

> Is it expected that the attentuation should work this differently in the two products?

I'm actually seeing bigger points with Joseph's parameters, though definitely not as dramatic as CesiumJS:
| Unity | CesiumJS |
| ----- | --------- |
| image | image |

But I should point out that in Sandcastle, the size of the canvas on my machine is considered 671.4px by 559.2px. I extracted these values from the "Elements" view in the Chrome debug window. This is compared to the 2015 x 888 pixel window I see in the Unity editor. Despite taking up more space on my screen, the Sandcastle window in Chrome is considered to be a very small canvas, so the points render much larger. I think the pixel size of the points in Unity is actually more accurate.

kring:
No luck on the AMD with these latest changes, @j9liu. The points seem to mostly disappear now when attentuation is enabled.
attenuation

kring:
Looks like the problem is that the VS is outputting the same 6 points over and over:
image

kring:
Copying the buffer also doesn't help anymore. When the Target of the new buffer is Raw, it looks exactly the same as the above (only a few points shown). If I set it to Structured, it looks very cool but not much like a church. :)

image

j9liu:
@kring ahh thanks for trying it out. I guess we need to copy the vertices into a structured buffer for DX11 after all 🙃 I'll undo my last commit and fix it tomorrow.

That does mean the points will take up 2x the space even with attenuation off, but I don't see a way to avoid that.

j9liu:
@kring updated! let me know if any other issues come up

kring:
Thanks @j9liu, great work!

PR #231, opened 2023-03-01
Check for successful Reinterop init before calling into C++ static methods

### LINKED ISSUE #230: Editor crashes trying to open scene with out of sync native code (`CesiumGlobeAnchor`)
Usually we can open the Unity Editor and interact with the scene, even if the native code isn't built. However, there's something that changed in #217 that is crashing the Editor on startup, instead of outputting a `NotImplementedException`:

```
=========================================================
	Native Crash Reporting
=================================================================
Got a UNKNOWN while executing native code. This usually indicates
a fatal error in the mono runtime or one of the native libraries 
used by your application.
=================================================================

=================================================================
	Managed Stacktrace:
=================================================================
	  at <unknown> <0xffffffff>
	  at CesiumForUnity.CesiumGlobeAnchor:DotNet_CesiumForUnity_CesiumGlobeAnchor_SetNewLocalToGlobeFixedMatrix <0x00094>
	  at CesiumForUnity.CesiumGlobeAnchor:SetNewLocalToGlobeFixedMatrix <0x0003a>
	  at CesiumForUnity.CesiumGlobeAnchor:UpdateEcef <0x000d2>
	  at CesiumForUnity.CesiumGlobeAnchor:Sync <0x003d2>
	  at CesiumForUnity.CesiumGlobeAnchor:Restart <0x000aa>
	  at CesiumForUnity.CesiumGlobeAnchor:OnEnable <0x0001a>
	  at System.Object:runtime_invoke_void__this__ <0x00087>
=================================================================
Received signal SIGSEGV
Obtained 0 stack frames
```

kring:
Fixes #230

Actually I'm not totally sure it does, because I can't reproduce #230, but I can see how it might happen. This PR makes sure reinterop is initialized before calling into any static C++ methods. If the C# and C++ code are out of sync, the initialization will throw an exception, rather than calling a function pointer using possibly the wrong signature and crashing.

### DIFF
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -18,11 +18,11 @@
 - `CesiumGlobeAnchor` now stores a precise, globe-relative orientation and scale in addition to position.
 - Added `localToGlobeFixedMatrix`, `longitudeLatitudeHeight`, `positionGlobeFixed`, `rotationGlobeFixed`, `rotationEastUpNorth`, `scaleGlobeFixed`, and `scaleEastUpNorth` properties to `CesiumGlobeAnchor`.
 - Added the `Restart` method to `CesiumGlobeAnchor`, which can be use to reinitialize the component from its serialized values.
-- Enabled caching of UnityWebRequests by copying all response headers.
+- Moved the Cesium tileset shaders from the `Shader Graphs` shader category to the new `Cesium` shader category.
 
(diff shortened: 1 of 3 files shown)


j9liu:
I think this worked, it's no longer crashing for me! 🎉

PR #239, opened 2023-03-06
libjpeg turbo

joseph-kaile:


kring:
Performance testing this PR with CesiumGS/cesium-native#604 merged into it, versus that branch by itself.

Cesium World Terrain in 01_CesiumWorld:

| *Refresh Number* | *Time (ms)* |
|---------------------|--------------|
| 1 | 715 |
| 2 | 780 |
| 3 | 877 |
| 4 | 815 |
| 5 | 782 |
| 6 | 774 |
| 7 | 634 |
| 8 | 823 |
| 9 | 911 |
| 10 | 696 |
| Min | 634 |
| Max | 911 |
| Median | 781  |
| Median reloads per hour | 4609 |

This should be directly comparable to the "after this PR" section of https://github.com/CesiumGS/cesium-native/pull/604#issuecomment-1463291001 and indicates about a 10% performance improvement!

kring:
Same test, but Melbourne Photogrammetry instead of Cesium World Terrain.

| *Refresh Number* | *Time (ms)* |
|------------------|--------------|
| 1 | 2346 |
| 2 | 1734 |
| 3 | 2331 |
| 4 | 3040 |
| 5 | 2611 |
| 6 | 3028 |
| 7 | 2036 |
| 8 | 2390 |
| 9 | 2782 |
| 10 | 2898 |
| Min | 1734 |
| Max | 3040 |
| Median | 2500.5 |
| Median reloads per hour | 1440 |

So that one is showing 13% slower if we look at the median. The min and max times are both faster with libjpeg-turbo, though (<1% and 6% respectively). The variability in load times of this tileset is so high that I really don't trust any it. Maybe I should look at that next.

Doing the same again:

| *Refresh Number* | *Time (ms)* |
|------------------|--------------|
| 1 | 1879 |
| 2 | 1964 |
| 3 | 2471 |
| 4 | 2114 |
| 5 | 2138 |
| 6 | 2552 |
| 7 | 2366 |
| 8 | 1894 |
| 9 | 1852 |
| 10 | 1933 |
| Min | 1852 |
| Max | 2552 |
| Median | 2039 |
| Median reloads per hour | 1766 |

And now the median is about 6.5% faster. Obviously this is all super suspect when the difference between the branches is much smaller than the difference between successive samples.

kring:
Time taken by the entire `readImage` method in this branch while loading the default Melbourne Photogrammetry view in Cesium for Unity:

image

Time taken by _just_ the STB section if I hack the method to use STB instead of libjpeg-turbo.

image

This is an impressive improvement, 2-3x faster!

PR #246, opened 2023-03-14
Support pre-existing mipmaps.

kring:
After CesiumGS/cesium-native#607, KTX2 textures still didn't work right in Cesium for Unity. The problem was that KTX2 provides the smallest mipmap first and the largest (highest resolution) last, while Unity expects the opposite.

This PR changes the texture loading code to read the mip levels from the correct place.


j9liu:
Thanks @kring , I confirmed this works! I'm going to merge it after #256 because of the conflict in `cesium-native`.

PR #252, opened 2023-03-20
Remove references to non-existent input axes

j9liu:
A forum user reported that the DynamicCamera threw runtime errors that said “Input Axis Controller Right Stick X is not setup” ([post here](https://community.cesium.com/t/runtime-error-input-axis-controller-right-stick-x-is-not-setup/22941/3)). This is because "Controller Right Stick X" and its Y counterpart are not built-in Unity variables. They appear in Unity's `FreeCamera.cs` script, but that script is actually creating the input mappings for that axis itself, which I didn't realize when I was referencing it for the input names.

There's no public Unity API to bind new inputs to `InputManager` at runtime. If a user wants to support camera movement with their controller, they'll have to create the input mapping itself. But this just involves duplicating the "Mouse X" and "Mouse Y" entries and configuring the duplicate entries for a controller joystick. See [here](https://www.reddit.com/r/Unity3D/comments/hgx88m/mouse_look_andor_right_stick_controller/) for more detail.

### DIFF
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -1,5 +1,11 @@
 # Change Log
 
+### v0.4.0
+
+##### Fixes :wrench:
+
+- Fixed a bug where `CesiumCameraController` tried to access non-existent inputs in the legacy input system.
+
 ### v0.3.1
 
 ##### Fixes :wrench:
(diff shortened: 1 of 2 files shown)


kring:
Thanks @j9liu!

PR #253, opened 2023-03-21
Fix shader graph to support built-in renderer

swardle:


kring:
Thanks @swardle, much appreciated!

PR #265, opened 2023-03-29
Proof of concept: Use Shader Graph for point cloud rendering 

### LINKED ISSUE #554: Add support for point cloud attentuation in the HDRP render pipeline
Since it was originally implemented, point cloud attentuation has only worked in the URP render pipeline. See note in the PR that added it:
https://github.com/CesiumGS/cesium-unity/pull/218#issuecomment-1460386474

This recently came up in a private support thread.

kring:
The point cloud rendering in #218 is awesome, but unfortunately it will only work with the Universal Render Pipeline. That's especially unfortunate because we've added support for the built-in pipeline in this release. I was curious what stopped us from implementing point cloud rendering using Shader Graph, which _should_ make it much easier (perhaps even automatic) to support all the render pipelines. So this draft PR is a proof of concept of doing that, and I don't see any major barriers to it. Short version: it works.

URP:
image

HDRP:
image

It uses a custom function node to read from the structured buffer, so no increase in memory usage. In fact, it should be pretty easy to use this same approach in the non-attenuated case, too, so we only need one copy of the points.

Because the output vertex position in shader graph is always in "object space", we need to transform the clip space position back to object space, only so Unity can go the other way again. I think this is the biggest downside to this approach. But GPUs basically do `matrix * vector` multiplications in their sleep, right? I haven't measured performance to be sure, but I think this is a worthwhile tradeoff for the compatibility that shader graph gives us.

It should also be possible to do nifty stuff like drape raster overlays over point clouds.

I'm currently using an "empty" `Mesh` to drive the rendering. It doesn't have any vertex data, but it does have index data because meshes require it. So this is pretty inefficient, but using a Mesh rather than DrawProcedural makes Unity set the model matrix and maybe other uniforms in the normal way that shader graph expects. We can definitely make this more efficient, I was just being lazy.

Stuff I haven't figured out (hopefully no deal killers here, but who knows?):

* [ ] How to conditionally compile custom HLSL in a custom function node based on shader keywords. Right now it's just hard-coded that the buffer has colors but no normals.
* [ ] Normals and Tangents. I've kind of ignored them and have the shader set to "unlit", because the lit shader doesn't work without vertex normals and tangents.
* [ ] For rendering to work in HDRP, I had to force the tileset to render opaque. Something in the transparent branch in UpdateMaterial breaks it and causes it to render nothing. (actually, looking in RenderDoc, it renders all the right triangles, but every fragment of them is fully transparent, so nothing is visible on the screen)
* [ ] The built-in render pipeline isn't working. Looking in RenderDoc, I can see the immediate problem is that the position coming out of the vertex shader is identical for every vertex, both within and across billboards. I don't know why, though.

@j9liu I'm curious what you think. I guess the first question is whether I've missed any major deal killers on this such that its not as practical as it appears. And the second question (assuming the answer to the first is positive) is whether we want to a) ship the current implementation in this release, and perhaps switch to shader graph next release, b) hold shipping attentuation until we sort this out, or c) try to wrap up this approach this week and get it into this release.

### DIFF
--- a/Runtime/CesiumPointCloudRenderer.cs
+++ b/Runtime/CesiumPointCloudRenderer.cs
@@ -36,6 +36,8 @@ namespace CesiumForUnity
 
         private Cesium3DTileInfo _tileInfo;
 
+        private Mesh _emptyMesh;
+
         public Cesium3DTileInfo tileInfo
         {
             set => this._tileInfo = value;
@@ -51,7 +53,7 @@ namespace CesiumForUnity
 
             this._pointCount = this._mesh.vertexCount;
             this._pointMaterial = UnityEngine.Object.Instantiate(
-                        Resources.Load<Material>("CesiumPointCloudShadingMaterial"));
+                        Resources.Load<Material>("CesiumPointCloudTilesetMaterial"));
 
             GraphicsBuffer sourceBuffer = this._mesh.GetVertexBuffer(0);
 
@@ -72,7 +74,7 @@ namespace CesiumForUnity
 
             if (this._mesh.HasVertexAttribute(VertexAttribute.Color))
             {
-                this._pointMaterial.EnableKeyword("HAS_POINT_COLORS");
+                this._pointMaterial.EnableKeyword("_HAS_POINT_COLORS");
             }
             else
             {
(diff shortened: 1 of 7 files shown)


j9liu:
I messed around with the shader graph today and the vertex ID node _does_ work in built-in. As proof, I removed the vertex deformations from the vertex shader and used this as the mesh color (the divisor is 100,000):



When I do `Graphics.DrawMesh` with the regular point mesh and the modified material, there's a clear gradient across the points:



Then I thought it had to do with retrieving the data from `StructuredBuffer`. Hardcoding `position` in `CesiumReadPointCloudData_float` to be `(0, 0, 0)`, then applying the XY offset based on vertexID in world space, gave me a bunch of horizontal planes, which means the vertexID definitely works:



And it actually turned out that the data _could_ be retrieved from the buffer if you hardcoded an integer. So when I add a `uint` cast, i.e. `uint pointIndex = (uint)vertexIndex / 6;`, it's able to retrieve the positions!



But enabling the rest of the code still makes the entire thing disappear, so there's still something else that's wrong. Have to find out where.

PR #269, opened 2023-03-30
Show credits in Editor UI

j9liu:
| On Screen Credits | With Popup | 
| ---------------------- | ---------- |
| image | image |

This PR modifies `CesiumCreditSystem` so that it can display credits in the scene view. Notably:
- The UI in `CesiumCreditSystem` has been put in its own file, `CesiumCreditSystemUI`.
- The UI now uses UIToolkit instead of UGUI (Unity's default runtime GUI)
- TextMeshPro will no longer be a required dependency.

TODOs:
- [x] Toggling "Show Credits On Screen" on a tileset doesn't refresh its credits.
- [x] Figure out life time management of `CesiumCreditSystem` between scenes and edit / play mode.


j9liu:
> Toggling "Show Credits On Screen" on a tileset doesn't refresh its credits.

I tried mocking this up real quick but I can't seem to get it right. Currently in `cesium-native` a tileset's "show credits on screen" option doesn't get properly propagated to the credits because the `Credit` constructor only checks for `html` equality, and doesn't do anything with the `showCreditsOnScreen`. I tried to fix this change, then add an event to `Cesium3DTileset` to refresh the credits, forcing it to override the `creditsUpdated` check we do to optimize frames without changes. But I can't sync up the actual moment that the tileset is reconstructed after `showCreditsOnScreen`, with the time that the credit system is updated. So when I did this, it always looked like the UI was out of sync with the actual option.

It seems like the most thorough solution would be to always iterate over the credits to detect those changes, and only broadcast the event update if the credits actually changed. I don't know how much that will impact performance, though.

j9liu:
This depends on https://github.com/CesiumGS/cesium-native/pull/617.

@kring do you mind taking a look at my approach? In summary, I always end up iterating over the credits in the scene, but I only parse their HTML / send the changes to the C# side if the credit counts actually changed. I know this adds overhead every frame, but it looks like CesiumJS has to empty / refill its credit lists every frame as well. Let me know what you think.

j9liu:
@kring this is ready for review. I changed the approach -- instead of iterating over the credits every frame, I added a `setShowCreditsOnScreen` function to `Tileset` in cesium-native. Then, whenever `showCreditsOnScreen` is toggled in the Unity `Cesium3DTileset`, I invoke an event that forces the credit system to update. This way, it accounts for changes to the setting without overhead every frame.

I also think I got it to work with `HideFlags.HideAndDontSave`. It required a lot of checks between environments to destroy the credit system, but with these checks I'm not seeing any visual or printed errors when I switch between scenes + go in and out of Play Mode. Let me know what you think.

kring:
I noticed that the credits weren't appearing at all in Play mode or in a built game. Using the UI Toolkit debugger, I learned that the credits existed, but were off the screen (at the top). I eventually noticed a warning about a missing runtime stylesheet when looking at the CesiumCreditSystemUIPanelSettings in the Inspector, saying the UI might not work right at runtime. Once I added a default stylesheet, the credits started working. I'm not quite sure why this didn't happen to you (different Unity version maybe?), but hopefully this is an ok change!

kring:
I used Unity's profiler to confirm that credit UI handling is an insignificant amount of frame time. Thanks @j9liu!

PR #272, opened 2023-04-03
Remove references to non-existent input axes (part 2)

j9liu:
It came up [on the forum](https://community.cesium.com/t/argumentexception-input-axis-yaxis-is-not-setup/23207) that _another_ axis that I thought existed didn't actually exist, even after #252 . This removes that final invalid axis, and I actually tested it this time to confirm there were no more runtime errors. 😅

### DIFF
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -1,5 +1,11 @@
 # Change Log
 
+### ? - ?
+
+##### Fixes :wrench:
+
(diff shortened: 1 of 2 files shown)


joseph-kaile:
Tested with the legacy input system. The Up/Down works a little differently, but it doesn't have errors.

PR #287, opened 2023-04-19
Generate flat normals

joseph-kaile:
This PR follows the Unreal code to generate flat normals. Also, adds support for Triangle Fan/Triangle Strip and fixes a bug in getting metadata when mesh primitive mode is Triangle Fan/Triangle Strip.


kring:
Thanks @joseph-kaile!

PR #289, opened 2023-04-21
Check triangle mesh before attempting to bake it for physics

j9liu:
This PR checks if a triangle mesh contains at least three vertices before trying to bake it into a physics mesh. This catches an error that can come from loading data with degenerate triangles: 

image

It's not a perfect solution, but at least it will prevent the app from crashing in play mode. Maybe we should `Debug.LogWarning` something about it?


kring:
Thank @j9liu!

PR #295, opened 2023-04-27
Fix Mesh memory leak when switching between play and edit modes

### LINKED ISSUE #293: Mesh pool results in memory leak between Editor and Play Modes
As noted on the [forum](https://community.cesium.com/t/cesium-for-unity-version-1-0-0-memory-exception/23660):

> I think it has to do with an object-pooling PR that made it into the 1.0.0 release: [Add object pooling by ZackOfAllTrades · Pull Request #244 · CesiumGS/cesium-unity · GitHub](https://github.com/CesiumGS/cesium-unity/pull/244). If I undo the changes made by that PR, the memory will reset every time it enters Play Mode. The mesh pool is expected to expand when you load more meshes, but it seems to happen even if I don’t move the camera around much. We’ll have to look into this to figure out what’s going on.

kring:
Fixes #293

There were two problems here:

First, the "get" callback and the "release" callback in the mesh object pool were reversed, meaning that the mesh data was cleared on _get_ rather than on _release_. As a result, meshes with big chunks of data would end up sitting in the Mesh object pool. This was easily fixed by flipping them.

However, even after this change, there was still a small-ish memory leak when switching between Play and Edit modes, caused by the Mesh objects themselves not being destroyed. The static `ObjectPool<Mesh>` - or the objects in it - were seemingly not being destroyed by whatever cleanup logic Unity executes at the end of Play mode.

I initially tried to add some manual cleanup, such as in the `DomainUnload` callback. But this only seemed to be called when entering Play mode, not when exiting it. And destroying Unity object's in this event would be pretty questionable, anyway. So then I tried (and failed) to find a more suitable event. `EditorApplication.playModeStateChanged` seemed promising, but was called too early. The in-use meshes were not yet released back into the pool by the time this event was raised. And Unity's ObjectPool seems to have a pretty obvious bug when calling `Clear` while objects associated with the pool are still in use. End result: broken pool, and we still had a memory leak.

I tried various ways to fix or work around this, but ended up just writing a less broken object pool myself. The pool is still disposed in `EditorApplication.playModeStateChanged`, but in the implementation here, a disposed object pool will simply destroy any objects that are released into it.

### DIFF
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -13,6 +13,7 @@
 - Added dependencies on the ShaderGraph and InputSystem packages to resolve material / script compilation errors.
 - Fixed another bug where `CesiumCameraController` tried to access a non-existent input in the legacy input system.
 - Removed an extra "delimiter" added to the end of on-screen credits in some cases.
+- Fixed a memory leak of `Mesh` objects when entering and exiting Play mode in the Unity Editor.
 
 ### v1.0.0 - 2023-04-03
 
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -2,6 +2,10 @@
 
 ### ? - ?
 
+##### Breaking Changes :mega:
+
+- `CesiumObjectPool` is no longer accessible from outside the CesiumRuntime assembly.
+
 ##### Additions :tada:
 
 - Added support for primitives with the `TRIANGLE_STRIP` and `TRIANGLE_FAN` topology types.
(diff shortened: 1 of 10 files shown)


j9liu (commented):
@kring looks good! I confirmed this removes the memory leak. There's just one fix for a typo but I can push a commit for that once CI passes

PR #297, opened 2023-04-27
fix primitive name

joseph-kaile:
When there are multiple primitives per mesh, the numbers become negative.


kring:
Thanks @joseph-kaile!

Issue #275, opened 2023-04-05
Changing a `CesiumGeoreference`'s transform at runtime causes tile displacement.


j9liu:
[Reported on the forum](https://community.cesium.com/t/adjusting-globe-orientation/23246/6). It seems like changes in the `CesiumGeoreference`'s transform aren't properly accounted for at runtime. This causes floating tiles to render:

![image](https://user-images.githubusercontent.com/32226860/230181225-224b688e-eef1-42bf-ac26-9593340953f5.png)

This may also be related to #132, though I haven't taken a close look at the math it's actually describing.

artoowang:
I think I have a method to reproduce this issue with Cesium 1.0.0. Please see the following commit for a test scene and repro steps:
https://github.com/artoowang/cesium-unity-samples/commit/7cc3d977953479277e09b9ad75668c880445b81d

The following video shows the issue (it was recorded in Unity 2022.2.2f1, in play mode, but the same issue also happens in Unity 2021.3.10f1 and in editor mode):
https://drive.google.com/file/d/13S1upEDgQVfoQy1v6mNpoOPC9flZZ4hj/view?usp=share_link

Also, there seems to be no issue with Cesium 0.2.0:
https://drive.google.com/file/d/1WihsQCHyXfDN9k2-kDUAX12U_jAjIMTn/view?usp=share_link

In the test above, it seems when I change the latitude, the system caches some new tiles. When I changed the latitude back, and moves the Georeference object instead, the previously cached tiles are shown, but at a wrong place.

kring:
Also reported here:
https://community.cesium.com/t/an-animation-problem-with-earth-scaling/23623

kring:
The problem occurs when both the georeference origin and a globe anchor's parent transform change in the same frame.

Because the parent transform has changed, and because we (incorrectly!) count the entire transform chain rather than just the local transform for transform change detection, the CesiumGlobeAnchor decides to recompute the ECEF matrix from the local transformation. This wouldn't be catastrophic - just introduce a bit of error, maybe - except for the fact that the georeference has also changed. So we end up recomputing the ECEF matrix from the Transform, but the Transform is only correct for the _old_ georeference, not the new one. So the tile moves.

[...]

PR #299, opened 2023-04-29
Changed CesiumGlobeAnchor to check local instead of the world transform

artoowang:
This fix is based on @kring 's suggestion at https://github.com/CesiumGS/cesium-unity/issues/275
Tested with the repro step in the issue discussions above, and all unit tests still pass.


artoowang:
Oops, sorry, I did something wrong. Please ignore this request.

PR #302, opened 2023-05-08
Only recompute the ECEF transformation if a GlobeAnchor's _local_ transformation changes

kring:
Fixes #275 

@artoowang this overlaps with your PR (#300). I had actually implemented at the same time I wrote the comment in #275, but was traveling and otherwise distracted and hadn't gotten around to opening a PR yet. I really appreciate you opening yours, though; you'll see it's very similar to this one. Once you have that CLA signed I'll merge yours as well, but I'm opening this one now so that we can get a version with this change shipped ASAP.

PR #303, opened 2023-05-08
Add the ability to scale the CesiumGeoreference

### LINKED ISSUE #280: Tilesets render black when scaled down in built-in render pipeline
This happens when you add a Cesium World Terrain w/ Bing Maps Imagery to the scene, then scale it down to 0.00006 in all three components (you may have to do so slowly, because the double-click to zoom function doesn't account for the new size).

This is a known issue based on the presence of multiple forum threads ([x](https://forum.unity.com/threads/textures-not-rendered-when-model-is-scaled-down.451451/), [x](https://forum.unity.com/threads/objects-with-small-scale-turn-black.1225566/), [x](https://forum.unity.com/threads/materials-turn-black-on-scaling.752732/)), but I don't know what the solution would be here. This issue on the [Mapbox Unity SDK](https://github.com/mapbox/mapbox-unity-sdk/issues/42) came up while I was searching for answers -- they suggest its a floating precision issue with the lighting system. That might be what's happening -- the tiles are so thin it's causing rendering artifacts.

### LINKED ISSUE #291: Very small scales cause jittering
In this video, the Melbourne tileset is inside a georeference with a scale of 1e-7 (making the entire globe roughly a meter across), and I'm just rotating the camera.

small-scale-jittering2

With this setup, the tile Transforms have relatively large (ECEF) positions, and the georeference has a tiny scale, and Unity is multiplying this transformation matrices together in single precision. We could do this multiplication in double-precision instead (though we'll have to fight Unity to do it), and that will probably help.

kring:
This a PR into #302 so merge that first.

Adds a `scale` property to `CesiumGeoreference` that can be used to precisely scale the entire globe. By doing the scaling computations in double precision, this offers more precision than adjusting the scale property on the Transform. I was able to scale the globe down to 1e-8 without any obvious vertex precision problems (though Unity's near and far plane limits become a bit of an issue at that scale).

Objects with a CesiumGlobeAnchor scale with the globe (as expected, hopefully).

Fixes #291

Unfortunately this does not fix #280. The built-in render pipeline still goes black suddenly when the scale is smaller than about 7e-05.

### DIFF
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -2,6 +2,10 @@
 
 ### ? - ?
 
+##### Additions :tada:
+
+- Added a `scale` property to `CesiumGeoreference`, allowing the entire globe to be scaled up or down while preserving precision better than adjusting the scale on the `Transform`.
+
 ##### Fixes :wrench:
 
 - Fixed a bug that caused primitive numbers to be negative in the names of tile game objects when the tile mesh had multiple primitives.
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -9,6 +9,7 @@
 ##### Fixes :wrench:
 
 - Fixed a bug that caused primitive numbers to be negative in the names of tile game objects when the tile mesh had multiple primitives.
+- Fixed a bug that caused tiles to be displaced when changing the transform of a `CesiumGeoreference` at runtime.
 
 ### v1.1.0 - 2023-05-01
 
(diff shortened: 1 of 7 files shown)


j9liu (commented):
@kring looks good! I just had one suggestion for changelog clarity.

I thought this change could have used a unit test, so I pushed a commit for one. If it looks sufficient, then this PR can be merged as you see fit.

PR #306, opened 2023-05-10
Fix missing textures when using multiple raster overlays with the same projection

kring:
This fixes the bug reported here:
https://community.cesium.com/t/tms-wms-raster-overlay-overlapping-is-flikering/23904

There was a bug in `UnityPrepareRendererResources::detachRasterInMainThread` where it would clear the wrong overlay texture slot in the material. Instead of using the index of the overlay in the list of overlays to look up the correct material texture slot, it would use the overlay texture coordinate index. With only one overlay, these are the same. With two overlays with different sets of texture coordinates (e.g., one web mercator the other geographic), they're also likely to be the same. But with two overlays with the same projection, these numbers are different, and attempts to detach the second texture would accidentally detach the first instead. End result: missing textures.

See the community forum post above for an easy way to test this. You can use `https://sandcastle.cesium.com/CesiumUnminified/Assets/Textures/NaturalEarthII/tilemapresource.xml` for the TMS URL rather than downloading it and loading it locally as described in the post.

### DIFF
--- a/native~/Runtime/src/UnityPrepareRendererResources.cpp
+++ b/native~/Runtime/src/UnityPrepareRendererResources.cpp
@@ -1474,7 +1474,6 @@ std::optional<uint32_t> findOverlayIndex(
     return std::nullopt;
 
   uint32_t overlayIndex = 0;
-  bool overlayFound = false;
   for (const CesiumUtility::IntrusivePointer<RasterOverlay>& pOverlay :
        pTileset->getOverlays()) {
     if (&overlay == pOverlay.get()) {


native~/Runtime/src/UnityPrepareRendererResources.cpp
```diff
@@ -1462,6 +1462,33 @@ void UnityPrepareRendererResources::freeRaster(
...
+    const UnityEngine::GameObject& tileset,
+    const Cesium3DTilesSelection::RasterOverlay& overlay) {
+  DotNet::CesiumForUnity::Cesium3DTileset tilesetComponent =
+      tileset.GetComponent<DotNet::CesiumForUnity::Cesium3DTileset>();
+  Tileset* pTileset = tilesetComponent.NativeImplementation().getTileset();
+  if (!pTileset)
+    return std::nullopt;
+
+  uint32_t overlayIndex = 0;
+  bool overlayFound = false;
```
j9liu, line 1477: Looks like this variable isn't being used, so I'll push a commit to delete it.

j9liu (commented):
@kring this looks good! I just had a comment that I can push a commit for.

PR #307, opened 2023-05-10
Detect material property changes in Editor mode

### LINKED ISSUE #180: Changing material parameters does not update material for all tiles

j9liu:
Fixes #180. This replaces the `Instantiate(material)` call with `MaterialPropertyBlock` instead, allowing for changes to the opaque material to be propagated to already-loaded tiles.

I did some initial profiling of the difference between `Material` and `MaterialPropertyBlock`. This is on the first sample scene, with just Cesium World Terrain and OSM Buildings. It's true that using the property block disables SRP batching, and it does have a noticeable performance difference: 

| `Material` | `MaterialPropertyBlock` |
| ---------- | --------------------------- |
| RegularMaterial | MaterialPropertyBlock |

In this scene, SRP batching allows us to stay just under 7ms per frame drawing. Without SRP batching, the drawing time fluctuates constantly between 9-12ms per frame.

Here's the performance on the second sample scene, with CWT and Melbourne Photogrammetry.

| `Material` | `MaterialPropertyBlock` |
| ---------- | --------------------------- |
| RegularMaterial2 | MaterialPropertyBlock2 |

For this scene, the `MaterialPropertyBlock` does increase the GPU time per frame, but not as drastically. And I'm not sure what makes Melbourne Photogrammetry more stable than other datasets?

Finally, I made a third sample scene with a view of San Francisco Photogrammetry and CWT.

| `Material` | `MaterialPropertyBlock` |
| ---------- | -------------------------- |
| RegularMaterial3 | MaterialPropertyBlock3 |

This looks more similar to the first sample scene's performance, though the difference is not as bad. There's fluctuation between 9-12ms per frame. San Francisco is probably not as "batchable" as Cesium OSM Buildings because it has more differences between tile material properties. Hence why the performance of OSM buildings is so much better with batching, but is relatively the same with the `MaterialPropertyBlock` approach.

Also, it seems like something is consistently generating 368B of GC every frame, but I can't imagine what. It doesn't seem to be related to this PR.

### DIFF
--- a/native~/Runtime/src/Cesium3DTilesetImpl.cpp
+++ b/native~/Runtime/src/Cesium3DTilesetImpl.cpp
@@ -470,6 +470,13 @@ void Cesium3DTilesetImpl::LoadTileset(
 
     excluder.AddToTileset(tileset);
   }
+
+  // If the tileset has an opaque material, set its hash here to avoid
+  // destroying it on the first tick after creation.
+  if (tileset.opaqueMaterial() != nullptr) {
+    int32_t opaqueMaterialHash = tileset.opaqueMaterial().ComputeCRC();
+    _lastOpaqueMaterialHash = opaqueMaterialHash;
+  }
 }
 
 } // namespace CesiumForUnityNative
--- a/native~/Runtime/src/Cesium3DTilesetImpl.cpp
+++ b/native~/Runtime/src/Cesium3DTilesetImpl.cpp
@@ -25,6 +25,7 @@
 #include <DotNet/UnityEngine/Camera.h>
 #include <DotNet/UnityEngine/Debug.h>
 #include <DotNet/UnityEngine/GameObject.h>
+#include <DotNet/UnityEngine/Material.h>
 #include <DotNet/UnityEngine/Quaternion.h>
 #include <DotNet/UnityEngine/Time.h>
 #include <DotNet/UnityEngine/Transform.h>
@@ -51,7 +52,8 @@ Cesium3DTilesetImpl::Cesium3DTilesetImpl(
       _updateInEditorCallback(nullptr),
 #endif
       _creditSystem(nullptr),
-      _destroyTilesetOnNextUpdate(false) {
+      _destroyTilesetOnNextUpdate(false),
+      _lastOpaqueMaterialHash(0) {
 }
 
 Cesium3DTilesetImpl::~Cesium3DTilesetImpl() {}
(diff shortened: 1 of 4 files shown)


j9liu:
I did attempt one more approach -- `Material.ComputeCRC` is a function that computes a hash for the contents of the material. In `Cesium3DTileset.Update()` I added a check that compared the opaque material's CRC to the last one, and if it was different (e.g., the base color changed), it would destroy the tileset. This does not have any affect on the GPU time per frame, and it seems to have a barely noticeable effect on the CPU time as well. Here's the profiler with the CPU time before and after adding an opaque material to Cesium World Terrain (the spike is the tileset recreating itself):

image

The downside is that whenever a user changes a property on the material, it has to refresh the entire dataset. This may not be ideal if a user wants to, say, toggle a tileset's color at runtime, without refreshing everything. I don't know what the answer is because I'm not sure what users will want most out of using materials? There is a performance penalty for using `MaterialPropertyBlock`, but a nice responsiveness that doesn't require the tileset to refreshed, and that might be a bigger win...

j9liu:
Based on offline discussion with @kring, we decided that the benefit of `MaterialPropertyBlock` doesn't justify the added performance overhead. On the other hand, refreshing the tileset every time a property is changed on a material may not be desirable. Outside of the Editor, this is undesirable for applications that dynamically update tileset materials.

We compromised on having the Editor refresh the material whenever the hash changes. But it is **expected behavior** that changing the opaque material's properties at runtime will **not** update existing tiles. It falls on the user to implement this in their own script. When they change a material's property, they'll have to iterate over all of the tileset's children and manually change the setting on those materials.

Here's the commit hash for the `MaterialPropertyBlock` approach, in case we ever want it again: https://github.com/CesiumGS/cesium-unity/pull/307/commits/18538079ad3ecb722b6efcbcbcccea7bd401a069

native~/Runtime/src/Cesium3DTilesetImpl.cpp
```diff
@@ -82,12 +84,24 @@ void Cesium3DTilesetImpl::Update(
...
+      return;
+    }
+
+    // If the opaque material or any of its properties have changed, recreate
+    // the tileset to reflect those changes.
+    if (tileset.opaqueMaterial() != nullptr) {
+      int32_t opaqueMaterialHash = tileset.opaqueMaterial().ComputeCRC();
+      if (_lastOpaqueMaterialHash != opaqueMaterialHash) {
+        this->DestroyTileset(tileset);
+        _lastOpaqueMaterialHash = opaqueMaterialHash;
```
kring, line 100: I think this needs to be set in LoadTileset, too. Otherwise it will always be destroyed on the first tick after it's first created.

kring (commented):
All makes sense to me! Just one small thing below.

kring (commented):
All makes sense to me! Just one small thing below.

j9liu:
@kring done! Thanks for the catch

PR #314, opened 2023-05-18
Remove now duplicated (tileset failure load) message

joseph-kaile:


kring:
Thanks @joseph-kaile!

Issue #190, opened 2023-01-27
Add "AddComponentMenu" attribute to Cesium scripts


kring:
Just an easy usability thing: we can add the [AddComponentMenu("Cesium")](https://docs.unity3d.com/ScriptReference/AddComponentMenu.html) attribute to our classes, like `Cesium3DTileset` and `CesiumGlobeAnchor`, to make them easy to find when clicking "Add Component" on a game object. There's also a second parameter that allows us to control the order in which the components appear.

PR #315, opened 2023-05-19
Add the "AddComponentMenu" attribute to all classes visible to the user

csciguy8:
Moves menu location from 'Scripts/Cesium for Unity' to 'Cesium'.

A few other visibility issues were spotted by @j9liu , but the original intent of this request should be satisfied. 

The cesium scripts are a little easier to find from the "Add Component" menus.

Fixes issue #190

### DIFF
--- a/Runtime/Cesium3DTileset.cs
+++ b/Runtime/Cesium3DTileset.cs
@@ -39,6 +39,7 @@ namespace CesiumForUnity
     /// </remarks>
     [ExecuteInEditMode]
     [ReinteropNativeImplementation("CesiumForUnityNative::Cesium3DTilesetImpl", "Cesium3DTilesetImpl.h")]
+    [AddComponentMenu("Cesium/Cesium 3D Tileset")]
     public partial class Cesium3DTileset : MonoBehaviour, IDisposable
     {
         public void Dispose()
--- a/Runtime/CesiumPointCloudRenderer.cs
+++ b/Runtime/CesiumPointCloudRenderer.cs
@@ -17,7 +17,7 @@ namespace CesiumForUnity
     }
 
     [ExecuteInEditMode]
-    [AddComponentMenu("Cesium/Cesium Point Cloud Renderer")]
+    [AddComponentMenu("")]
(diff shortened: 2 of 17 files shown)


Runtime/CesiumPointCloudRenderer.cs
```diff
@@ -17,6 +17,7 @@ internal struct Cesium3DTileInfo
     }
 
     [ExecuteInEditMode]
+    [AddComponentMenu("Cesium/Cesium Point Cloud Renderer")]
```
j9liu, line 20: This component is internal and meant only to be added by tilesets. I'm surprised it shows up in the component menu despite its C# class visibility.

Giving the attribute an empty string hides the component from the list. Let's do this here to hide this from users.

```suggestion
    [AddComponentMenu("")]
```

Runtime/Cesium3DTileset.cs
```diff
@@ -39,6 +39,7 @@ public enum CesiumDataSource
     /// </remarks>
```
j9liu, line 39: Since this is a user-facing change, I think it's worth mentioning in `CHANGES.md`. Could you update it to mention this change, following the format of the other change log entries?

Under the "Additions" section, this can say something like: Cesium components now appear under the "Cesium" category in the Component menu.

j9liu (commented):
Congratulations on your first PR @csciguy8 ! 🎉 This looks great overall, just a few comments before it can be merged.

The main thing is that `CesiumPointCloudRenderer` shouldn't be addable by a user (and I'm surprised it was addable even before this PR). `CesiumCreditSystem` and `CesiumCreditSystemUI` are probably irrelevant to the user too, but it's not as problematic if they show up here, so don't worry about it.

Runtime/Cesium3DTileset.cs
csciguy8: Done.

j9liu:
@csciguy8 looks great, thanks! Usually we wait for CI to complete but since this is a UI-focused change, I'm happy to merge now. Congrats on your first contribution to Cesium! 🎉

Issue #191, opened 2023-01-27
Add "IconAttribute" to Cesium scripts


kring:
https://docs.unity3d.com/ScriptReference/IconAttribute.html

This should let us use the Cesium logo or whatever we want as the icon for our script classes, like `Cesium3DTileset`, rather than the generic `#` script icon:

![image](https://user-images.githubusercontent.com/924374/215073932-6104c4f4-7c36-4300-bf5a-61a1c140cc7e.png)

PR #316, opened 2023-05-19
Add icon attributes to cesium scripts

csciguy8:
For all user facing Cesium scripts, instead of the using the default Unity script icon, use the Cesium logo.

A new 24x24 Cesium icon was created to match the rendering area of the icons in the Inspector and Add Component menu.

Resolves issue #191


csciguy8:
@kring Kevin, do you have an opinion about scripts in the "Runtime" folder accessing resources in the "Editor" folder?

That's what this change is currently doing. It could be argued that the icon should be in the "Runtime/Resources" folder for organizational consistency, but in the end, it's an editor specific tag anyway, and none of this may really matter.

kring:
Hmm yeah it's a good question. It does seem a little odd at first glance to have something in Runtime reference Editor. I tried to find examples of how others have structured their projects, and found that `com.unity.2d.animation` uses the same pattern you've used here. Others seem to utilize the fact that a simple string like `MyIcon` will (apparently) be resolved to `MyIcon.png` in any `Resources` directory anywhere. But using a full path seems safer.

All of which is a long way to say: this looks good to me. :)

kring:
Just one request: please add a note to the Additions section of `CHANGES.md` mentioning that we've added icons.

csciguy8:
Thanks for the insight Kevin. Also, I've added a note to `CHANGES.md` like you've requested.

j9liu:
I'll go ahead and merge this. Thanks @csciguy8 !

Issue #175, opened 2023-01-12
Changing `CesiumGeoreference` while a sub-scene is active does not change the sub-scene's origin


j9liu:
When a `CesiumSubScene` is active, any changes to the sub-scene origin will be reflected on the `CesiumGeoreference`. But this is not true the other way around. When the `CesiumGeoreference` origin is changed, the active `CesiumSubScene` origin should be changed to the same values.

This would just require changes to `CesiumGeoreference.UpdateOrigin()`. You'd have to iterate over the sub-scenes in the `CesiumGeoreference`'s children, finding the active one (only one can be active at a time), and set its origin coordinates to be the same as the `CesiumGeoreference`'s.

PR #320, opened 2023-05-24
Georef subscene origin sync

csciguy8:
Let any changes to `CesiumGeoreference` origin propagate to the active subscene, if exists.
Refactored `CesiumSubScene` to register itself as a listener to any parent `CesiumGeoreference` changes (`changed ` delegate). Similar to the pattern that `CesiumGlobeAnchor` uses.
A special `OnParentGeorefChanged` function ensures that parent changes don't propagate even more parent changes.
Added 2 new unit tests.

Resolves Issue #175


csciguy8:
Done and done, @kring . Thanks for the review.

kring:
Thanks @csciguy8!

PR #326, opened 2023-06-01
Fix cracks

### LINKED ISSUE #325: Cracks between tiles in Melbourne tileset
As reported here:
https://community.cesium.com/t/3dtilestileset-show-s-the-problem/23840/4

These cracks don't exist in Cesium for Unreal so I don't think it's a problem with the tileset. I also don't think they used to exist in Cesium for Unity, even, so we likely introduced them recently. Possibly in #278 or #287?

joseph-kaile:
Fixes #325

This problem was fixed by copying logic from Cesium for Unreal. (https://github.com/CesiumGS/cesium-unreal/blob/ue5-main/Source/CesiumRuntime/Private/CesiumTextureUtility.cpp). If I change the code to always generate mipmaps, the cracks also appear in Unreal.


kring:
Thanks @joseph-kaile!

PR #331, opened 2023-06-06
fix missing zlib library causing xcode build failure

joseph-kaile:


kring:
Thanks @joseph-kaile!

PR #336, opened 2023-06-20
Added event to notify when new GameObject is created.

joseph-kaile:
In trying to integrate TeleportationArea from the XR Interaction Toolkit, the GameObject/colliders need to be registered with the XR Interaction Manager. An easy way to do this is to simply add a TeleportationArea script to the GameObject when its created.

With the new event, it can easily be achieved:
```
cesium3DTileset.OnNewGameObjectCreated += go  => go.AddComponent<TeleportationArea>();
```

### DIFF
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -1,5 +1,11 @@
 # Change Log
 
+### ? - ?
+
+##### Additions :tada:
+
+- Added an event to notify when a new GameObject is created.
+
 ### v1.3.1 - 2023-06-06
 
 ##### Fixes :wrench:
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -4,7 +4,7 @@
 
 ##### Additions :tada:
 
-- Added an event to notify when a new GameObject is created.
+- Added `OnTileGameObjectCreated` event to `Cesium3DTileset` class, which allows customizing the Tile GameObjects as they are loaded.
 
 ### v1.3.1 - 2023-06-06
 
--- a/Runtime/Cesium3DTileset.cs
+++ b/Runtime/Cesium3DTileset.cs
@@ -70,6 +70,16 @@ namespace CesiumForUnity
             }
         }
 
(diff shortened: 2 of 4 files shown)


Runtime/Cesium3DTileset.cs
```diff
@@ -70,6 +70,16 @@ internal static void
             }
         }
 
+        public event Action<GameObject> OnNewGameObjectCreated;
```
j9liu, line 73: Maybe the name should be more specific to the tileset, like `OnTileGameObjectCreated`? @kring @csciguy8 do you have any strong opinions?

j9liu, line 73: Add comments for this event, similar to the others. Something like this:
```
/// <summary>
/// An event that is raised when a new tile game object is created for the tileset.
/// This can be used to modify tiles as they are loaded -- for example, adding additional components.
/// </summary>
```

CHANGES.md
```diff
@@ -1,5 +1,11 @@
 # Change Log
 
+### ? - ?
+
+##### Additions :tada:
+
+- Added an event to notify when a new GameObject is created.
```
j9liu, line 7:
```suggestion
- Added an event to notify when a new GameObject is created for a tile in a `Cesium3DTileset`.
```

j9liu (commented):
@joseph-kaile I confirmed this works, just have a few suggestions for naming and documentation.

Can you also merge with main to fix the CHANGES.md conflict? Thanks!

joseph-kaile:
> @joseph-kaile I confirmed this works, just have a few suggestions for naming and documentation.
> 
> Can you also merge with main to fix the CHANGES.md conflict? Thanks!

Thanks for the suggestions.

j9liu:
Looks great @joseph-kaile ! I'll just confirm that CI passes before merging to main. Thanks!

PR #337, opened 2023-06-21
Change occlusion value of '0' to mean full occlusion.

joseph-kaile:
This was causing shadows to be extremely dark. 
Before and after:
before
after

In both Unity and glTF,  Occlusion of 0.0 means full occlusion, 1.0 means fully lit:

glTF:
occlusion : The occlusion texture; it indicates areas that receive less indirect lighting from ambient sources. Direct lighting is not affected. The red channel of the texture encodes the occlusion value, where 0.0 means fully-occluded area (no indirect lighting) and 1.0 means not occluded area (full indirect lighting). Other texture channels (if present) do not affect occlusion.
The texture binding for occlusion maps MAY optionally contain a scalar strength value that is used to reduce the occlusion effect. When present, it affects the occlusion value as **1.0 + strength * (occlusionTexture - 1.0).**

[Unity ](https://docs.unity3d.com/Packages/com.unity.render-pipelines.high-definition@10.10/manual/Ambient-Occlusion.html):
When authoring ambient occlusion Textures, be aware that a value of 0 specifies an area that is fully occluded and a value of 1 specifies an area that is fully visible.

Also, Shadergraph and Surface Shaders have a default occlusion value 1.0 which means no ambient occlusion.

So, by using the equation above:

new-ambient-occlusion

### DIFF
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -1,5 +1,11 @@
 # Change Log
 
+### ? - ?
+
+##### Fixes :wrench:
+
+- Fixed how the occlusion strength is used in the default tileset shader which was causing shadows to be too dark.
+
 ### v1.3.1 - 2023-06-06
 
 ##### Fixes :wrench:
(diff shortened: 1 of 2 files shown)


j9liu:
Looks great @joseph-kaile, thanks!

PR #338, opened 2023-06-27
Record prefab anchor modification when Transform is modified.

### LINKED ISSUE #335: Globe Anchor Doesn't Support Prefab Overrides
If the `CesiumGlobeAnchor` component is on a GameObject in a prefab, the prefab's world coordinates will always be used regardless of coordinate overrides for prefab instances.

For example, I have a prefab of a helicopter with a world anchor:
image
It has a default coordinate of `33.4588218, -111.72937073` because that's where I originally made it a prefab.

If I place multiple instances of the prefab, all instances will revert back to this coordinate either when "Play" mode enters, or the scene is reloaded:
image

Two helicopters now overlapping at the same coordinate after play/reload:
image

Notice how `Scale` has a blue highlight showing its an override, while the globe anchor has nothing marked as changed:
image

I fixed the issue for now by not including the anchor in the prefab, and I manually add the anchor to each instance.

kring:
Fixes #335

### DIFF
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -1,5 +1,11 @@
 # Change Log
 
+### ? - ?
+
+##### Fixes :wrench:
+
+- Fixed a bug that caused a prefab with a `CesiumGlobeAnchor` to lose its position after save/reload.
+
 ### v1.3.1 - 2023-06-06
 
 ##### Fixes :wrench:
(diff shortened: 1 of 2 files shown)


j9liu:
Looks great, thanks @kring !

PR #345, opened 2023-07-06
Added bounds property for tileset

JeroenTekle:
I've added a bounds property to Cesium3DTileset. Personally I use this to teleport to tiles added manually by users. 

The bounds calculating code is copied & modified from `Cesium3DTileImpl::getBounds`. The tileset needs some time to load before this property can be used. `_pTileset` needs to be set & `getRootTile()` needs to be available as well. Because of this I now return bounds with all 0's but I don't think this is the most elegant solution. How would you suggest to handle this?

### DIFF
--- a/Runtime/Cesium3DTileset.cs
+++ b/Runtime/Cesium3DTileset.cs
@@ -650,6 +650,17 @@ namespace CesiumForUnity
             }
         }
 
+        /// <summary>
+        /// Gets the axis-aligned bounding box of the root tile of this tileset.
+        /// </summary>
+        public Bounds bounds
+        {
+            get
+            {
+                return getBounds();
+            }
+        }
+
         private partial void SetShowCreditsOnScreen(bool value);
 
         private partial void Start();
@@ -669,5 +680,7 @@ namespace CesiumForUnity
         /// Zoom the Editor camera to this tileset. This method does nothing outside of the Editor.
         /// </summary>
         public partial void FocusTileset();
+
+        private partial Bounds getBounds();
     }
 }
--- a/native~/Runtime/src/Cesium3DTilesetImpl.cpp
+++ b/native~/Runtime/src/Cesium3DTilesetImpl.cpp
@@ -4,6 +4,7 @@
 #include "UnityPrepareRendererResources.h"
 #include "UnityTileExcluderAdaptor.h"
 #include "UnityTilesetExternals.h"
+#include "UnityTransforms.h"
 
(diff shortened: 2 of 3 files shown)


JeroenTekle:
@kring is this something of interest for this project or not?

kring:
Sorry this got by me @JeroenTekle. This seems like a useful feature to me. Based on a quick review though, I think some work needs to be done on the coordinate system. Currently the bounds will be returned as an AABB in ECEF, which is probably not what users need or expect. It should be transformed to the Cesium3DTileset game object's coordinate system, and the documentation of the Bounds property should say as much. You'll notice that `FocusTileset` does a lot more work around coordinate transformation.

Also, before we can merge this, we do need you to sign the Contributor License Agreement if you haven't already:
https://github.com/CesiumGS/cesium/blob/master/CONTRIBUTING.md#contributor-license-agreement-cla

JeroenTekle:
No problem @kring. 

For my specific use case I actually need coordinates in ECEF (or GCS) to be able to set the CesiumGeoreference's Origin to the tileset center. I could however convert the game object's local coordinates bounds using `CesiumGeoreference::TransformUnityPositionToEarthCenteredEarthFixed`.

In this case it might actually be better to expose the root `Cesium3DTile` through the `Cesium3DTileset` and use it's `bounds` property instead? I'm not sure however how to accomplish casting the root tile `Cesium3DTilesSelection::Tile*` returned by `Cesium3DTilesSelection::Tileset::getRootTile()` to a `DotNet::CesiumForUnity::Cesium3DTile`. How would I go about this?

kring:
> In this case it might actually be better to expose the root Cesium3DTile through the Cesium3DTileset and use it's bounds property instead?

That's not unreasonable, but lifetime management will be tricky. You can simply construct a `Cesium3DTile` instance from the native code, and then set its `_pTile` property with a pointer to the native Tile instance. See `UnityTileExcluderAdaptor.cpp`. The trouble is that Tile might later be destroyed, and when it is, trying to use the Cesium3DTile will cause a hard crash. In the case of the tile excluder, we get around this (more or less) by saying in the docs that you're not allowed to hold a Cesium3DTile reference past the end of the ShouldExclude call that provided it. If the root tile is available as a Cesium3DTile property on Cesium3DTileset, there's no obvious valid scope. It's valid until the root tile is destroyed, which will happen on many (but not all) changes to Cesium3DTileset properties.

So I think just providing a way to get the Bounds directly will be simpler for now. But you can definitely implement that by constructing a Cesium3DTile and then getting and returning its bounds.

Issue #327, opened 2023-06-01
Add support for multiple cameras


kring:
Currently, Cesium for Unity _exclusively_ uses [Camera.main](https://docs.unity3d.com/ScriptReference/Camera-main.html) for tile selection. We need to provide users with a way to add more cameras when needed. Something like Cesium for Unreal's CameraManager is a good catch-all, but there are probably some easier/automatic things we can do as well, such as collecting all cameras tagged `MainCamera`.

Requested here: https://community.cesium.com/t/about-more-main-camera-in-unity/24433/4
And here: https://community.cesium.com/t/get-location-or-bounds-of-tileset/25177/5

jonmartincom:
I found this Main Cam issue as load time seemed awful, I.e. not loading from nearest point but filling from the far back frustrum, so changed my AR cams to not be main and used the main cam for a top down view, rendering AR views of the content loaded from other cams, it worked 10-20x faster than having it try and load in a full horizontal type view! good call though although I think your talking about the opposite problem of only using one cameras view over many.

dhyams:
I have run into this issue as well, although I am a Unity newbie and might be misusing a camera.  In my experience,  if a second camera is added, the landscape in general does not fill in, on one or the other of the cameras.  I believe that this is the same problem as described in the original posting by @kring.

<img width="1455" alt="image" src="https://github.com/CesiumGS/cesium-unity/assets/1141744/96efeb8b-223b-4360-8f5e-4d21ec8b2311">

kring:
Yes that sounds like the same problem @dhyams.

dhyams:
I was looking and trying to fix this; and it seems pretty clear that (at least at first implementation) all it would take would be to iterate through all cameras and add ViewStates to the returned array

https://github.com/CesiumGS/cesium-unity/blob/3bd1519845f92a236727c93810e92936a51fe51a/native~/Runtime/src/CameraManager.cpp#L85

But, I don't see any API for getting all cameras on the C++ side, equivalent to https://docs.unity3d.com/ScriptReference/Camera-allCameras.html

I was wondering if anyone knew how to go get a list of cameras on the C++ side; I am sure I'm missing something.  But I'd like to help.

kring:
Making new Unity functionality available to C++ is easy. See here:
https://cesium.com/blog/2023/01/26/developing-for-unity/

In short, you just need to add some code that uses `Camera.allCameras` to the `ConfigureReinterop.cs` file and let Unity compile it.

[...]

PR #352, opened 2023-08-04
Add support for Universal Windows Platform (UWP) / Holo Lens 2

kring:
It supports both ARM64 and x64 processors. UWP apps can be run on normal Windows and distributed on the Windows Store, as well.

Some notes / setup instructions:

1. Add "Universal Windows Platform Build Support" to your Unity version via the Unity Hub.
2. Add the "Universal Windows Platform Development" workload to Visual Studio using its installer.
3. Add the "C++ (v143) Universal Windows Platform tools" optional component. (if you're using VS2019 it will be v142 instead of v143)
4. The "Architecture" dropdown in the Universal Windows Platform Build Settings doesn't seem to actually matter. The native code is build for both x64 and ARM64. 32-bit platforms aren't supported. The Visual Studio solution generated by Unity can be used to build for either platform, regardless of what was selected here.
5. Be sure that the `InternetClient` Capability is enabled under Project Settings -> Player -> Universal Windows Platform tab -> Capabilities. Otherwise, we won't be able to load tilesets from the internet.

### DIFF
--- a/native~/CMakeLists.txt
+++ b/native~/CMakeLists.txt
@@ -1,4 +1,4 @@
-cmake_minimum_required(VERSION 3.15)
+cmake_minimum_required(VERSION 3.18)
 set(CMAKE_EXPORT_COMPILE_COMMANDS ON)
 
 project(CesiumForUnityNative
@@ -28,6 +28,19 @@ set(CMAKE_POSITION_INDEPENDENT_CODE ON)
 set(BUILD_SHARED_LIB OFF)
 set( SUPPORT_CONSOLE_APP OFF)
 add_subdirectory(extern/tidy-html5 EXCLUDE_FROM_ALL)
+
+# Patch tidy-html5's sprtf.c to work on UWP. UWP doesn't have lstrlen, but
+# it's totally unnecessary because simply passing -1 as the length will do the same thing.
+# So just swap in the -1.
+if (${CMAKE_SYSTEM_NAME} MATCHES "WindowsStore")
+  get_target_property(tidySources tidy-static "SOURCES")
+  list(REMOVE_ITEM tidySources "src/sprtf.c")
+  file(READ "extern/tidy-html5/src/sprtf.c" sprtSource)
+  string(REPLACE "int len = (int)lstrlen(ps);" "int len = -1;" sprtSource "${sprtSource}")
+  file(CONFIGURE OUTPUT "${CMAKE_CURRENT_BINARY_DIR}/extern/tidy-html5/sprtf.c" CONTENT "${sprtSource}")
+  list(APPEND tidySources "${CMAKE_CURRENT_BINARY_DIR}/extern/tidy-html5/sprtf.c")
+  set_target_properties(tidy-static PROPERTIES "SOURCES" "${tidySources}")
+endif()
 
 if (${CMAKE_SYSTEM_NAME} STREQUAL "Android")
   set(HTTPLIB_USE_OPENSSL_IF_AVAILABLE OFF)
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -24,6 +24,9 @@ jobs:
           submodules: recursive
       - name: Install nasm
         uses: ilammy/setup-nasm@v1.4.0
+      - name: Install Visual Studio UWP support
+        run: |
+          choco install -y visualstudio2022-workload-universal
       - name: Install Unity UWP support
         run: |
           start -FilePath "C:\Program Files\Unity Hub\Unity Hub.exe" -ArgumentList "-- --headless install-modules --version 2021.3.13f1 --changeset 9e7d58001ecf --module universal-windows-platform uwp-il2cpp" -Wait
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -26,7 +26,7 @@ jobs:
         uses: ilammy/setup-nasm@v1.4.0
       - name: Install Visual Studio UWP support
         run: |
-          choco install -y visualstudio2019-workload-universal --params "--add Microsoft.VisualStudio.ComponentGroup.UWP.VC"
+          choco install -y visualstudio2019-workload-universal --params "--add Microsoft.VisualStudio.ComponentGroup.UWP.VC --add Microsoft.VisualStudio.Component.Windows11SDK.22000"
       - name: Install Unity UWP support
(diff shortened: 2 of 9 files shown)


native~/CMakeLists.txt
```diff
@@ -1,4 +1,4 @@
-cmake_minimum_required(VERSION 3.15)
+cmake_minimum_required(VERSION 3.18)
```
kring, line 1: This is needed because of the use of `FILE(CONFIGURE)` below. CMake 3.18 came out July 2020, so hopefully this isn't a problem.

csciguy8 (commented):
The code changes look good. 

I've also tested the CI artifact locally, using it to build a sample project to Windows, UWP Intel64, and UWP Arm64. I was unable to test the Arm64 build (no hardware).

Some follow up questions:
1) For your setup instructions, should we put this in the developer-setup.md? Or even somewhere in the Quickstart tutorials?
2) Same questions for the other platforms. I see references to the platforms we support [here](https://cesium.com/platform/cesium-for-unity/), but no technical documentation or helpful hints.
3) Following up on that, installing the "C++ (v143) Universal Windows Platform tools" optional component tripped me up. I installed a similar "individual component" that did not work. And worse, Unity gave me a misleading error message to fix it
`Burst requires Visual Studio 2017 or greater (in addition, the following Visual Studio components need to be installed via the Visual Studio Installer: 
		(1) the "Universal Windows Platform development" workload, 
		(2) "C++ Universal Windows Platform support for vNNN build tools (ARM64)", 
		(3) "MSVC vNNN - VS NNNN C++ ARM build tools (Latest)", and
		(4) "MSVC vNNN - VS NNNN C++ ARM64 build tools (Latest)") in order to build a standalone player for UWP`
Maybe instructions with screenshots would help that? Or maybe I just like to get bit by MSVS dependency problems.

kring:
> For your setup instructions, should we put this in the developer-setup.md? Or even somewhere in the Quickstart tutorials?

I think it could make sense to have a tutorial for HoloLens specifically at some point. Written by someone who has tried it on an actual device, preferably. But in the meantime, Unity's UWP instructions cover most of it (including some bits I didn't mention above): https://docs.unity3d.com/Manual/WindowsStore.html

The only moderately Cesium-specific notes to add to that are: 1) we only support 64-bit platforms, and 2) InternetClient permission is required if you're trying to load tilesets from the internet. (1) is mentioned on our releases page (and will be updated with the new platform support when we release it). (2) is hopefully obvious? In any case, I'm not sure where would be most useful to put notes like that. The HoloLens tutorial, yes. Maybe a Unity FAQ?

But I don't think this needs to get wrapped up in this PR.

> Same questions for the other platforms. I see references to the platforms we support [here](https://cesium.com/platform/cesium-for-unity/), but no technical documentation or helpful hints.

Ok, but definitely not part of this PR.

> Following up on that, installing the "C++ (v143) Universal Windows Platform tools" optional component tripped me up. I installed a similar "individual component" that did not work. And worse, Unity gave me a misleading error message to fix it

Maybe this was a result of following my quick notes rather than Unity's much more detailed instructions for building for this platform?

csciguy8:
All sensible comments. I think all my concerns are something for the future, when Hololens or UWP becomes a more major platform of concern.

paulio:
I've followed the basic Cesium Unity Tutorials up to placing an object on the Melbourne photogrammetry tile. I've then added the MRTK scene components and turned off the Cesium Dynamic camera's flight controller. I'm going to investigate further, but any hints, such as confirming the SolidColor as unknown is helpful.

kring:
I have zero HoloLens experience, so these suggestions will be generic at best. But some things to try:

1. Try other tilesets. Melbourne is extremely heavy from a texture perspective, so perhaps it's hitting a weakness of the device. Cesium World Terrain + Bing and the Google Photorealistic 3D Tiles would be good ones to try.
2. In fact, try, the Cesium for Unity Samples project rather than creating your own project from scratch by following the tutorial.
3. The Samples project has a VR sample scene. You're not doing VR exactly, but still some of the configuration in that scene might be valuable on HL as well.
4. Try using the built-in renderer instead of URP or HDRP. We've seen much better performance with built-in on Oculus, at least.
5. Increase the Max SSE value to a higher value. This will result in less detail, but it may be needed given the limitations of the device.
6. You didn't answer my question about using a build from our CI system. But if you're not, and instead you built it yourself, make sure you're using a Release build. This makes a massive difference in performance.

paulio:
Sorry, I used an early UWP supported release that pointed me here, just says 1.5.0. The Skybox issue was a setting in the MRTK Display Settings, it overrides the Background Color AND clip ranges. In the editor that has tripled the FPS performance...now for the headset.

paulio:
FYI I need to carry out more investigation. If I disable the Cesium components the FPS is still very low, or rather it quickly drains to become very low. I'll try and narrow down what's happening.

paulio:
As recommended, I brought in the New York City VR scene and exported the CesiumGeoreference node. I created a new project and added MRTK v2 to the scene, i.e. lost all the XR locomotion, pointers, etc. Ensured that the camera was set to black in the MRTK setting and Release mode was used for the final build to the HL. That enabled me to run the example at an okay 20 FPS. Switching to Bing Maps version of New York also worked, a little slower and also hit memory exceptions after a couple of mins. So in summary it does work, sort of. My next puzzle is that I want Cesium to be displayed at a fixed point, e.g. on a table. The problem is that both Cesium and MRTK seem to search for CameraMain. Is there a known way to have one camera rendering the Cesium aspect without it being Main?

j9liu:
Hi @paulio,

There is currently no way to specify a different camera other than Main. We're aware of this inconvenience, and we have an issue open for it at #327.

In the meantime, if there's anything else we can help with, feel free to post to the [community forum](https://community.cesium.com/)! We reserve Github for confirmed bugs and feature requests, and are better able to provide support on the forum. Plus, there may be other forum users who are experiencing similar issues that can view the same thread.

paulio:
I've got a little further on. By using `Microsoft.MixedReality.Toolkit.Utilities.CameraCache.UpdateCachedMainCamera(cameraToBecomeMain);` MRTK now uses a different camera to that marked as main. That allows for the separation of cameras. The problem that's left is that the Cesium OSM Buildings do not seem to be respecting the culling. 
image

raggnic:
@paulio did you get further?

PR #356, opened 2023-08-16
Add ComputeLoadProgress to C# Cesium3DTileset

joseph-kaile:
Added ComputeLoadProgress function to estimate the percentage of the 3D tileset that has been loaded for the current view.

Example code on how to compute the time it takes for a tileset to load:

```using UnityEngine;
using CesiumForUnity;
using System.Diagnostics;

public class PerformanceTesting : MonoBehaviour
{
    Cesium3DTileset _tileset;
    public UnityEngine.UI.Text text;
    Stopwatch stopwatch = Stopwatch.StartNew();
    long elapsedTimeInMilliseconds = 0;
    bool finished = false;

    private void Start()
    {
        _tileset = this.GetComponent<Cesium3DTileset>();
    }
    void Update()
    {
        float loadProgress = _tileset.ComputeLoadProgress();
        if(loadProgress != 100.0f)
        {
            finished = false;
        }
        else if(!finished && loadProgress == 100.0f)
        {
            finished = true;
            elapsedTimeInMilliseconds = stopwatch.ElapsedMilliseconds;
        }
        text.text = "Time taken to load: " + elapsedTimeInMilliseconds + "\n";
    }
}
```

### DIFF
--- a/native~/Runtime/src/Cesium3DTilesetImpl.cpp
+++ b/native~/Runtime/src/Cesium3DTilesetImpl.cpp
@@ -328,10 +328,10 @@ void Cesium3DTilesetImpl::FocusTileset(
 
 float Cesium3DTilesetImpl::ComputeLoadProgress(
     const DotNet::CesiumForUnity::Cesium3DTileset& tileset) {
-      if(getTileset() == nullptr){
-        return 0;
-      }
-      return getTileset()->computeLoadProgress();
+  if (getTileset() == nullptr) {
+    return 0;
+  }
+  return getTileset()->computeLoadProgress();
 }
 
 Tileset* Cesium3DTilesetImpl::getTileset() { return this->_pTileset.get(); }
--- a/native~/Runtime/src/Cesium3DTilesetImpl.cpp
+++ b/native~/Runtime/src/Cesium3DTilesetImpl.cpp
@@ -326,6 +326,14 @@ void Cesium3DTilesetImpl::FocusTileset(
 #endif
 }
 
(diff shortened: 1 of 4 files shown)


native~/Runtime/src/Cesium3DTilesetImpl.cpp
```diff
@@ -326,6 +326,14 @@ void Cesium3DTilesetImpl::FocusTileset(
...
 }
 
+float Cesium3DTilesetImpl::ComputeLoadProgress(
+    const DotNet::CesiumForUnity::Cesium3DTileset& tileset) {
+  if (getTileset() == nullptr) {
+    return 0;
+  }
+  return getTileset()->computeLoadProgress();
+}
+
```
csciguy8, line 336: I think generally, as a helper function, this is fine. 

But I'm wondering if it's more something that the caller function should implement, and we should leave out of this code base. Here's why:
* The caller already has access to getTileset()
* This function is small and trivial to implement
* This function is implemented in a specific way. It's interpreting a nullptr tileset as 0 progress. The calling code might want this state handled in a different way.

Open for discussion for sure. I lean towards "less is more". Less to maintain, less to go wrong, etc.

kring: As Joseph mentioned offline, this functionality is trivial for us to implement in our C++ code using Reinterop, but it's impossible for users to implement themselves from C# (or even from C++ without modifying the plugin itself). So it's definitely valuable and needed.

kring:
Thanks @joseph-kaile!

PR #358, opened 2023-08-18
Add missing check for ETC1

joseph-kaile:
Fixed a bug where KTX tilesets did not display properly on iOS devices due to a missing check for ETC1 texture format.


kring:
Thanks @joseph-kaile!

PR #361, opened 2023-09-01
Convert to generic Resources.Load calls for Texture2D

kalvinpearce:
Swapped out the manual casts in the `Resources.Load` calls for their generic equivalent so that Unity won't try to pick an invalid `Texture2D` with the same name.

This case arose when using the [com.unity.vectorgraphics](https://docs.unity3d.com/Packages/com.unity.vectorgraphics@2.0/manual/index.html) package. As I understand it, the Vector Graphics package overwrites the import type of `.svg` files and they are no longer able to be cast to `Texture2D`. This was causing a few of the Cesium editors to throw exceptions and not render anything to the inspector. By using the generic version of `Resources.Load` Unity will pick valid `Texture2D` (in this case, the `.png` files rather than the `.svg`).

### DIFF
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -1,5 +1,11 @@
 # Change Log
 
+### ? - ?
+
+##### Fixes :wrench:
+
+- Fixed a bug that prevented editor windows from functioning when `com.unity.vectorgraphics` package was installed.
+
 ### v1.6.0
 
 ##### Additions :tada:
--- a/Editor/CesiumEditorStyle.cs
+++ b/Editor/CesiumEditorStyle.cs
@@ -45,7 +45,7 @@ namespace CesiumForUnity
 
         private static Texture2D LoadImage(string resourcePath)
         {
-            Texture2D icon = (Texture2D)Resources.Load(resourcePath);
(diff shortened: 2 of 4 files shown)


j9liu:
Thanks for the PR @kalvinpearce! Before we review, can you ensure that:

1. you have signed a CLA with us, either [individual](https://docs.google.com/forms/d/e/1FAIpQLScU-yvQdcdjCFHkNXwdNeEXx5Qhu45QXuWX_uF5qiLGFSEwlA/viewform) or [corporate](https://docs.google.com/forms/d/e/1FAIpQLSeYEaWlBl1tQEiegfHMuqnH9VxyfgXGyIw13C2sN7Fj3J3GVA/viewform)
2. you have updated `CHANGES.md` with a short description of the change? This sounds like it would go under the "Fixes" section, since it fixes bugs when using Cesium for Unity with `com.unity.vectorgraphics`.

kalvinpearce:
🙇🏼 I can confirm I have now signed the CLA and updated `CHANGES.md`. Please let me know if you'd like me to adjust any wording or formatting of the changelog file.

Issue #357, opened 2023-08-16
Check if XCode Project can build iOS app in CI


joseph-kaile:
The current CI process only verifies if Unity can export the XCode project, but not if the XCode project can build the iOS app. This has caused problems when adding external projects (Zlib and meshoptimizer) that need to be included in the Unity CMakeFile. The issue is to find a way for CI to detect and report any errors in the XCode project before users try to build the iOS app on their Macs.

kring:
Fixed in #363.

PR #363, opened 2023-09-04
Build XCode project as part of the build process

### LINKED ISSUE #362: v1.6.1 fails to build on iOS (again)
As reported here:
https://community.cesium.com/t/xcode-cesuim-version-1-6-build-errors-w-unity-2021-3-11/26519

This looks like another case of forgetting to add a new dependency to the `ALL_TARGETS` list in `native~/CMakeLists.txt`. Easy to fix, but we really need a way to detect this at CI time because it has happened a lot and most of the team has no way to test for it manually.

joseph-kaile:
Fixes #362 and #357

As part of the iOS building process, build the XCode project and ensure no errors. This ensures the iOS plugin from CI is not missing any libraries.

### DIFF
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -1,5 +1,11 @@
 # Change Log
 
+### ? - ?
+
+##### Fixes :wrench:
+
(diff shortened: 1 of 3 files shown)


kring:
Thanks @joseph-kaile!

PR #365, opened 2023-09-13
Fix Mac Standalone

joseph-kaile:
This code fixes the issue here https://community.cesium.com/t/missing-dlls-in-osx-standalone-il2cpp-builds/26714.


kring:
Thanks @joseph-kaile!

PR #373, opened 2023-10-24
Add Google to the Quick Add panel

kring:


csciguy8 (approved):
Code looks reasonable and matches the equivalent changes in cesium-unreal.

I'll leave it to you to verify full functionality from an account that has this asset enabled.

PR #386, opened 2023-12-06
Fix generated code problems with nullable primitive return values

kring:
As a test, I added the following to `Runtime/Helpers.cs`:

```csharp
public static bool? TestThing()
{
    return null;
}
```

And the following to `Runtime/ConfigureReinterop.cs`:

```csharp
Helpers.TestThing();
```

Reinterop generated invalid code for this. After this PR, the code is generated correctly. We previously handled nullable blittable structs, but not nullable primitives.


j9liu:
This works well on my `upgrade-feature-metadata` branch so I'm going to merge. Thanks @kring !

PR #389, opened 2024-01-11
Refactor API for 3D Tiles 1.1 metadata support

j9liu:
This PR refactors the plugin API to handle 3D Tiles 1.1 metadata implementation, which allows us to finally upgrade to the newest version of Cesium Native.

**TODO:**
- [ ] Finish writing unit tests for `CesiumMetadataValue`
- [ ] Look over for any missing functions between Unity / Unreal versions
- [ ] Get this to compile with clang
- [ ] Verify that this works in VR

Other:
- I can't write tests for most of the classes because they require a mock glTF struct to be initialized in C++. The tests don't have their own Reinterop so a test class would have to be included in the Runtime package itself. Should I do that?
- I didn't add any vector -> string or matrix -> string conversions because if we want to match Unity's string conventions, I'd have to write more code to handle this. I can do this if we want, but wasn't sure if it was a worthy effort.


j9liu:
I think this PR can finally be reviewed! Though it is a meaty one 😬

native~/Runtime/src/TestGltfModelImpl.cpp
```diff
@@ -0,0 +1,657 @@
...
+}
+
+DotNet::CesiumForUnity::CesiumFeatureIdTexture
+TestGltfModelImpl::AddFeatureIdTexture(
+    const DotNet::CesiumForUnity::TestGltfModel& model,
+    const DotNet::System::Array1<std::uint16_t>& featureIds,
+    const int64_t featureCount,
+    const DotNet::System::Array1<DotNet::Unity::Mathematics::float2>& uvs) {
+  const int32_t featureIdsLength = featureIds.Length();
+  assert(length == 4);
```
kring, line 86: This line causes a compiler error in debug builds:
```
error C2065: 'length': undeclared identifier
```

Runtime/CesiumIntMatN.cs
```diff
@@ -0,0 +1,395 @@
+using System.Diagnostics;
+using System;
+using Unity.Mathematics;
+
+namespace CesiumForUnity
+{
+    /// <summary>
+    /// Represents a mat2x2 with signed integer components. This preserves the accuracy of the integer
+    /// values, which could otherwise lose precision if represented as doubles.
```
kring, line 9: FWIW, only 64-bit integers might lose precision. All 8, 16, and 32-bit integers can be represented perfectly as doubles.

Runtime/CesiumMetadata.cs
```diff
@@ -7,10 +8,16 @@ namespace CesiumForUnity
...
     [ReinteropNativeImplementation("CesiumForUnityNative::CesiumMetadataImpl", "CesiumMetadataImpl.h")]
-    [AddComponentMenu("Cesium/Cesium Metadata")]
+    [AddComponentMenu("")]
     [IconAttribute("Packages/com.cesium.unity/Editor/Resources/Cesium-24x24.png")]
+    [Obsolete("Retrieve metadata using the CesiumModelMetadata component attached to a tile instead.")]
     public partial class CesiumMetadata : MonoBehaviour
     {
+        private void OnEnable()
+        {
+            Debug.LogWarning("CesiumMetadata component is deprecated. Retrieving metadata using the CesiumModelMetadata component of a tile object.");
```
kring, line 18:
```suggestion
            Debug.LogWarning("CesiumMetadata component is deprecated. Retrieve metadata using the CesiumModelMetadata component of a tile object.");
```

Runtime/CesiumMetadataValue.cs
```diff
@@ -0,0 +1,1825 @@
...
+        #endregion
+
+        #region Constructors
+        public CesiumMetadataValue() : this(null)
+        { }
+
+        public CesiumMetadataValue(System.Object value)
+        {
+            // For consistent implementation, convert intN to CesiumIntVecN, uintN to CesiumUintVecN,
+            // intNxN to CesiumIntMatNxN, and uintNxN to CesiumUintMatNxN
```
kring, line 62: In what way does that help the implementation be more consistent? Most likely I just need to keep reading, but the advantage is really not obvious to me at this point.

kring (commented):
Just getting started here, but things look great so far. A couple of comments below. I'm also wondering if you have any existing scenes you've been using to test various metadata configurations? Or examples of how to work with metadata from scripts?

Runtime/CesiumFeatureIdAttribute.cs
```diff
@@ -0,0 +1,66 @@
...
+        public override partial Int64 GetFeatureIdForVertex(Int64 vertexIndex);
+
+        /// <inheritdoc />
+        public override Int64 GetFeatureIdFromRaycastHit(RaycastHit hitInfo)
+        {
+            int vertex = CesiumPrimitiveFeatures.GetFirstVertexFromHitTriangle(hitInfo);
+            return this.GetFeatureIdForVertex(vertex);
+        }
+
+        internal override void CallDispose()
```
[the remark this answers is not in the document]
j9liu: `CesiumFeatureIdSet` isn't `IDisposable` because it doesn't store any native data, but its children have native data and therefore need to be disposed. Though... I suppose I can make `CesiumFeatureIdSet` disposable as well, and just have `Dispose` do nothing

Runtime/CesiumMetadataValue.cs
j9liu: There's no `Int8`, `Int16`, or `Int64` counterparts to `intN`, so I wrote the `CesiumIntVecN` struct to hold these values as well as their component type. I thought it would be a bit more inconvenient to deal with both `intN` and `CesiumIntVecN` cases, so I just moved the `intN` values into a `CesiumIntVecN` with `Int32` component type.

Runtime/CesiumIntMatN.cs
j9liu: That's good to know! I still prefer having `CesiumIntMatN` because it stores a component type, indicating whether its values are meant to be interpreted as `Int8`, `Int16`, `Int32`, or `Int64`, and that gets lost if represented as a Unity `dmat2x2`. But I'll modify the comment.

j9liu:
Thanks @kring ! I just opened https://github.com/CesiumGS/cesium-unity-samples/pull/65 in the Unity samples with changes based on the new API.

Also, you can test feature ID texture picking with San Francisco Ferry Building, asset 2333904. The script should work without needing to be changed.

Runtime/CesiumFeatureIdAttribute.cs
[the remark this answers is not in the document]
kring: Yeah it's pretty common for a base class to be disposable even though the base itself doesn't have anything to dispose. It's all about setting a contract.

kring (commented):
Another epic PR @j9liu! It looks great, I just noticed a few small things.

native~/Editor/CMakeLists.txt
```diff
@@ -46,6 +46,12 @@ set_target_properties(
         CXX_EXTENSIONS NO
 )
 
+target_compile_options(
+  CesiumForUnityNative-Editor
+  PRIVATE
+    -bigobj
```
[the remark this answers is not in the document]
j9liu: Ah yeah, when I realized had it had to be wrapped in the `if` statement I changed it back to `/bigobj` for the `Runtime` file, and forgot to do it for this one. 🤦

Runtime/CesiumModelMetadata.cs
```diff
@@ -0,0 +1,26 @@
...
+    /// <see cref="Cesium3DTileset"/>. It holds views of property tables
+    /// available on the glTF.
+    /// </summary>
+    /// <remarks>
+    /// This component is automatically added 
+    /// to tile game objects if their models contain the root extension.
+    /// </remarks>
+    [IconAttribute("Packages/com.cesium.unity/Editor/Resources/Cesium-24x24.png")]
+    [AddComponentMenu("")]
+    public class CesiumModelMetadata : MonoBehaviour
```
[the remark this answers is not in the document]
j9liu: No, not yet. I'm happy to add it later, but for the scope of this PR I intended to keep things functionally the same while upgrading `cesium-native`.

j9liu:
Thanks @kring ! I forgot to mention that I addressed all the comments and this is ready for another look 😅

kring:
Looks great @j9liu! 🚀

PR #391, opened 2024-01-18
Modify CesiumSubScene to use parent CesiumGeoreference's coordinates when added in-editor.

### LINKED ISSUE #343: A new CesiumSubLevel should adopt the currently-active georeference origin
Currently, when you create add a new CesiumSubLevel component to a GameObject, it is given the default georeference origin (near Denver) and then switches the currently-active georeference to that origin. This isn't great, because the user most likely wanted to create a sub-level where they were already looking, not in Denver. Fortunately, the "Create Sub-Level Here" is an effective workflow workaround, but we should still make it behave better when a sub-level is created manually.

azrogers:
Fixes #343.

When `CesiumSubScene::OnEnable` is called, it sets the coordinates of the parent `CesiumGeoreference` to the coordinates it has set. This means when adding a new `CesiumSubScene` component as a child of an existing `CesiumGeoreference`, it would reset the coordinates of the parent to the default coordinates. This change uses the `Reset` event to restore the georeference's original coordinates without changing the behavior outside of the editor.

### DIFF
--- a/Runtime/CesiumSubScene.cs
+++ b/Runtime/CesiumSubScene.cs
@@ -191,6 +191,16 @@ namespace CesiumForUnity
         [NonSerialized]
         internal CesiumGeoreference _parentGeoreference;
 
+        // The coordinates of the parent CesiumGeoreference before being changed by `UpdateOrigin`.
+        // This is because `OnEnable` runs before `Reset` does, so to be able to use the parent's
+        // coordinates as the default values without changing the OnEnable behavior of CesiumSubScene,
+        // we need to store what those values were before.
+        [NonSerialized]
+        private double3 _oldParentCoordinates = double3.zero;
+
+        [NonSerialized]
+        private CesiumGeoreferenceOriginAuthority _oldParentOriginAuthority;
+
         /// <summary>
         /// Sets the origin of the coordinate system to particular <see cref="ecefX"/>, <see cref="ecefY"/>,
         /// <see cref="ecefZ"/> coordinates in the Earth-Centered, Earth-Fixed (ECEF) frame.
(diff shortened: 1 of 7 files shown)


Runtime/CesiumSubScene.cs
```diff
@@ -310,13 +347,7 @@ private void OnParentChanged()
...
             // Update our origin to our parent georef, maintain our origin authority,
             // and copy both sets of reference coordinates. No need to calculate any of this again
-            this._longitude = this._parentGeoreference.longitude;
-            this._latitude = this._parentGeoreference.latitude;
-            this._height = this._parentGeoreference.height;
-
-            this._ecefX = this._parentGeoreference.ecefX;
-            this._ecefY = this._parentGeoreference.ecefY;
-            this._ecefZ = this._parentGeoreference.ecefZ;
+            CopyParentCoordinates();
```
j9liu, line 350: Just a nitpick for stylistic consistency.

```suggestion
            this.CopyParentCoordinates();
```

j9liu (commented):
Thanks @azrogers ! I hadn't realized that `OnEnable` runs before `Reset`, so I can see how this became more complicated than anticipated.

Still I wonder if this can be simplified (ish)? Instead of storing the parent coordinates, is it possible to have a `bool` variable called `isInitialized`? (name can be improved if possible)

Then in `OnEnable`, you could have this code towards the end
```
if (Application.isPlaying)
  // In play mode, the subscene coordinates should be used as-is.
  this.isInitialized = true;

if (isInitialized) {
 this.UpdateOrigin()
 }
 ```
 
Then `Reset` could look something like this:
```
this.UpdateParentReference();
if(this._parentGeoreference != null) {
  // adopt parent georeference coordinates here
  isInitialized = true;
}
```

How does that sound? If it helps with clarity, we could potentially surround these segments with `#if EDITOR` statements.

I confirmed that this implementation works otherwise. I just think the internal state management is a little confusing.

j9liu:
Actually, the solution in my previous comment fails to account for existing sub-scenes when they're enabled in the editor. 🙃 So ignore what I wrote. I think the solution you found is fine; I'd like to add a bit of clarity to the comment but I'll put a suggestion tomorrow morning.

Just because it was driving me crazy, I looked up if there was an event for any time a component was added in the editor. I found that the [ObjectFactory](https://docs.unity3d.com/ScriptReference/ObjectFactory.html) class exists _and_ has a `componentWasAdded` event on it, but I have no idea when this event is called in relation to the component's event cycle. It's probably not worth exploring at this point.

azrogers:
@j9liu Updated review based on feedback - let me know if I missed anything!

j9liu:
Looks great, thanks @azrogers !

PR #393, opened 2024-01-19
Fix warning when adding new DynamicCamera in the editor

### LINKED ISSUE #321: Console warning when adding a Dynamic Camera
Install Unity Hub 3.4.2 and Unity 2021.3.25f1.

Follow the 'Cesium for Unity Quickstart" guide,
https://cesium.com/learn/unity/unity-quickstart/

1. When creating the new project, use the "3D Sample Scene (URP)" template.
2. When installing Cesium from the Package Manager, use "Cesium for Unity, Version 1.2.0 - May 09, 2023"
3. Continue with the tutorial until you add a Dynamic Camera.

In the Console Window, you'll see this warning:

`DynamicCamera(Clone) is not nested inside a game object with a CesiumGeoreference. Most of its CesiumGlobeAnchor functionality will not work.
`

If you look at the Hierarchy Window, the Dynamic Camera is indeed under a CesiumGeoreference, which makes this puzzling. 
Have I done something wrong? Is the dynamic camera I created not going to work?

(note: Also happens when using the "3D Sample Scene (HDRP)" template)

azrogers:
Fixes #321.

When adding a new DynamicCamera from the Cesium editor utility, it would pop up a warning that the camera wasn't parented to a `CesiumGeoreference` despite this being the case. By using the overload of `Object.Instantiate` that supports specifying the new object's parent transform, the DynamicCamera's parent will be set by the time it initializes, avoiding this warning.

### DIFF
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -5,6 +5,7 @@
 ##### Fixes :wrench:
 
 - Fixed a bug that prevented the default `CesiumIonServer` asset from remembering its token in a clean project.
+- Fixed "DynamicCamera is not nested inside a game object with a CesiumGeoreference" warning when adding a new DynamicCamera in the editor.
 
 ### v1.7.0 - 2023-12-14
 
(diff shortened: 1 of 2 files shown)


j9liu:
Thanks @azrogers !

Issue #334, opened 2023-06-08
Deprecate CesiumFlyToController::flyToGranularityDegrees


csciguy8:
Rework `CesiumFlyToController` to incorporate the same improvements made to cesium-unreal's `AGlobeAwareDefaultPawn`.
- Change flight logic to not use segmented approximation, but refer to the source ellisoid directly (improved accuracy)
- Fix altitude calculation to incorporate geodetic normal, instead of geocentric (improved accuracy)
- Fix potential performance hiccups caused with very small `flyToGranularityDegrees` values

This work has been done in cesium-unreal, and can be applied the cesium-unity as well.
https://github.com/CesiumGS/cesium-unreal/pull/1120

We may also consider moving shared functionality between these two plugins to cesium-native. (@kring )

Issue #380, opened 2023-11-17
Update to 3D Tiles 1.1 metadata


j9liu:
We should update the metadata system to handle the 3D Tiles 1.1 extensions, instead of the deprecated `EXT_feature_metadata`. 
Currently, we have a single `CesiumMetadata` component that enables retrieving features from a particular triangle in a tile, which is really different from Cesium for Unreal. I think Cesium for Unreal's API is much more flexible, so I've been trying to replicate it in Unity, down to the same classes and structs.

My idea was to attach `CesiumModelMetadata` components to the game objects representing glTF models, and `CesiumPrimitiveFeatures` to the game objects representing primitives. @kring expressed concern about performance. I also worried this might not be scalable, because I experienced some lag while developing. However... my laptop was not plugged in, so that skewed my perception of things 😅 

I tested the new implementation more seriously today, and found that the difference isn't actually that noticeable. I did two views on the 05_CesiumMetadata level; first, the initial starting view, and second, a horizon shot containing many more buildings.

| View 1 | View 2 |
| ------- | --------- |
| ![image](https://github.com/CesiumGS/cesium-unity/assets/32226860/528cddf4-c0a4-4e1f-b5b2-a3171df11349) | ![image](https://github.com/CesiumGS/cesium-unity/assets/32226860/c248dc06-4005-4689-b891-3fec76f835f6) |

Using the performance profiler, I looked at the `Update.ScriptRunBehaviorUpdate` look under the function overview, since it tracked how long each component's `Update` took in a frame.

**View 1:**
| Single Component | Multi Component | 
| -------------------- | -------------------- |
| ![old](https://github.com/CesiumGS/cesium-unity/assets/32226860/f6127d21-d06e-4ef1-8ae6-93fab65b08a6) | ![image](https://github.com/CesiumGS/cesium-unity/assets/32226860/bb1acebb-193f-46b2-a807-d31bc26fec5e) |

**View 2:**
| Single Component | Multi Component | 
| -------------------- | -------------------- |
| ![Old2](https://github.com/CesiumGS/cesium-unity/assets/32226860/ad4dd786-cf85-4235-b9ca-b6e9e947913f) | ![image](https://github.com/CesiumGS/cesium-unity/assets/32226860/26c2402e-bbaa-4400-8de7-d13d86f44357)|

[...]

PR #395, opened 2024-01-23
Rewrite CesiumFlyToController to match improvements in cesium-unreal

### LINKED ISSUE #390: CesiumFlyToComponent "jumps" at the end due to spherical math that doesn't quite work on an ellipsoid
Copied from CesiumGS/cesium-unreal#1332.

Here's a diagram to try to illustrate the problem:
image

The blue oval is a highly exaggerated ellipsoidal model of the Earth. To do a flight between the start and end points marked, `CesiumFlyComponent` first computes a rotation axis and angle between those two points. At each step, it interpolates the angle and rotates the initial direction by that angle to produce the direction of the path at this step. It then scales this direction to the ellipsoid surface and adds a separately-computed height to it to get the actual position for the flight.

The problem, as shown in the diagram, is that scaling the final direction to the ellipsoid and then adding the final position's height does not actually result in the final position. It would on a sphere, but not on an ellipsoid. It's our old friend geocentric versus geodetic angles.

As a result, flights often appear to jump to a new position just as they complete, because the final position along the curve is quite different from the real final position. This can be seen quite obviously by opening the Google Photorealistic 3D Tiles level in Cesium for Unreal Samples, hitting Play-In-Editor, pressing 3 to fly to Tokyo, then moving the camera a good distance _up_ and away from Tokyo before pressing 3 to fly toward it again. When the intial position of the camera is fairly high, the jump at the end should be very obvious.

It's probably easy to fix this, but I'm not sure it's worth the trouble. I'm inclined to just port the camera flight code from CesiumJS rather than trying to fix this approach.

I originally wrote this issue for Cesium for Unreal, but Unity has the same problem. However it may be less obvious in Unity because we generate keypoints and then interpolate over them. So the "jump" will happen between the second-to-last and last keypoints rather than very suddenly at the end.

azrogers:
Fixes #390 and #334.

The `CesiumFlyToController` had a jump at the end of its flight path, similar to the bug in the Unreal plugin. However, the fix that we used for the Unreal plugin couldn't just be dropped in as the way the two implementations work was quite different. This change rewrites `CesiumFlyToController` to match the implementation of the Unreal equivalent, fixing the jump bug in the process.

### DIFF
--- a/Runtime/CesiumFlyToController.cs
+++ b/Runtime/CesiumFlyToController.cs
@@ -1,7 +1,6 @@
 using System;
-using System.Collections.Generic;
-using UnityEngine;
 using Unity.Mathematics;
+using UnityEngine;
 
 namespace CesiumForUnity
 {
@@ -81,7 +80,7 @@ namespace CesiumForUnity
             set => this._flyToDuration = Math.Max(value, 0.0);
         }
 
-        [SerializeField]
+        #region Deprecated Functionality
         [Min(0.0f)]
         private double _flyToGranularityDegrees = 0.01;
 
@@ -95,6 +94,7 @@ namespace CesiumForUnity
         /// The lower the value, the more keypoints are generated, and the smoother the flight
         /// interpolation will be.
         /// </remarks>
+        [Obsolete("CesiumFlyToController no longer works using keypoints. This value has no effect.")]
         public double flyToGranularityDegrees
         {
             get => this._flyToGranularityDegrees;
@@ -110,6 +110,7 @@ namespace CesiumForUnity
                 }
             }
         }
+        #endregion
 
         /// <summary>
         /// Encapsulates a method that is called whenever the controller finishes flying.
@@ -136,12 +137,21 @@ namespace CesiumForUnity
 
         private CesiumCameraController _cameraController;
 
-        private List<double3> _keypoints = new List<double3>();
-        private double3 _lastPositionECEF = new double3(0.0, 0.0, 0.0);
+        private double _sourceHeight;
+        private Vector3 _sourceDirection;
+        private Quaternion _sourceRotation;
+
+        private double3 _destinationECEF;
+        private Quaternion _destinationRotation;
+        private double _destinationHeight;
+
+        private Vector3 _rotationAxis;
+        private double _totalAngle;
+        private double _maxHeight;
+
+        private double3 _previousPositionECEF;
 
         private double _currentFlyToTime = 0.0;
-        private Quaternion _flyToSourceRotation = Quaternion.identity;
-        private Quaternion _flyToDestinationRotation = Quaternion.identity;
 
         private bool _flyingToLocation = false;
         private bool _canInterruptFlight = true;
@@ -185,7 +195,7 @@ namespace CesiumForUnity
         private bool DetectMovementInput()
         {
             double3 currentPositionECEF = this._globeAnchor.positionGlobeFixed;
-            bool3 positionEquality = currentPositionECEF == this._lastPositionECEF;
+            bool3 positionEquality = currentPositionECEF == this._previousPositionECEF;
             return !positionEquality.x || !positionEquality.y || !positionEquality.z;
         }
 
@@ -210,7 +220,12 @@ namespace CesiumForUnity
         /// <param name="deltaTime"> The time delta in seconds.</param>
         private void HandleFlightStep(float deltaTime)
         {
-            if (this._georeference == null || this._keypoints.Count == 0)
+            if (!this._flyingToLocation)
+            {
+                return;
+            }
+
+            if (this._georeference == null)
             {
                 this._flyingToLocation = false;
                 return;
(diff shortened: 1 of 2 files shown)


Runtime/CesiumFlyToController.cs
```diff
@@ -222,62 +237,67 @@ private void HandleFlightStep(float deltaTime)
...
-            this._currentFlyToTime += (double)deltaTime;
+            this._currentFlyToTime += deltaTime;
 
-            // If we reached the end, set actual destination location and orientation
+            double flyPercentage = 0.0;
             if (this._currentFlyToTime >= this._flyToDuration)
+            {
+                flyPercentage = 1.0;
+            }
+            else if (this._flyToProgressCurve != null && this._flyToProgressCurve.length > 0)
```
csciguy8, line 247: I like the length check here. This doesn't exist on the Unreal version, which might be a nice improvement.

csciguy8 (changes requested):
Code looks good to me when comparing against the latest Unreal implementation.

General functionality look ok, but my testing on unity samples, 04_CesiumSubScenes, showed that for small distances, I can see a small jump between at the end of the animation. (move away from the start a bit, then press 1)

Runtime/CesiumFlyToController.cs
```diff
@@ -222,62 +237,67 @@ private void HandleFlightStep(float deltaTime)
...
-            double flyPercentage = percentage;
-            if (this._flyToProgressCurve != null && this._flyToProgressCurve.length > 0)
-            {
-                flyPercentage = Math.Clamp(
-                    (double)this._flyToProgressCurve.Evaluate((float)percentage),
-                    0.0,
-                    1.0);
-            }
+            // Rotate the source direction around the axis by the given percentage of the total angle we're rotating, giving us our current position.
+            double3 rotatedDirection = (float3)(Quaternion.AngleAxis((float)(flyPercentage * this._totalAngle), this._rotationAxis) * this._sourceDirection);
```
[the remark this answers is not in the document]
azrogers: Nope! Originally `rotatedDirection` was a `Vector3`, so you need to use `float3` as an in-between to allow casting to a `Vector3`, but it turns out it made more sense as a `double3`. I'll remove this.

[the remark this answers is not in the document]
azrogers: Nevermind, now I remember why I did it - making `rotatedDirection` a `double3` avoids a cast in the next line, and a `Vector3` has to be cast to a `float3` to be assigned to a `double3`.

azrogers:
Updated review! @csciguy8. This one is pretty hefty. It's now blocked by CesiumGS/cesium-native#797 - the cesium-native implementation of `GlobeFlightPath` - as well as #380, which is required to get this project up to date with the latest version of cesium-native. Because of that, it'll probably take a while to get this one merged!

With that being said, I've moved the math behind `CesiumFlyToController` onto the cesium-native side so we can take advantage of glm's math functions to get the precision we need. More info about that side of things is in the cesium-native issue linked in the paragraph above. Then there's the glue in the cesium-unity native plugin to connect things up between C++ and Unity.

Speaking of which, I'd like some input from @kring on that part if possible. `CeisumGlobeFlightPath` isn't supposed to be constructed directly - you're supposed to go through `FromEarthCenteredEarthFixedCoordinates` or `FromLongituteLatitudeHeight`. The reason for this is because the creation of the `GlobeFlightPath` can fail (if the coordinates couldn't be mapped to the WGS84 geodetic surface) and needs to be able to return an empty value (the C++ side uses std::optional for this). But it doesn't look like Reinterop supports making a non-public constructor for a class, meaning you can happily call `new CesiumGlobeFlightPath` and you won't know you've done something wrong until an assertion on the C++ side crashes Unity. How can I solve this? Is there a different paradigm I should use for constructing objects where that construction can fail?

kring:
> But it doesn't look like Reinterop supports making a non-public constructor for a class, meaning you can happily call new CesiumGlobeFlightPath and you won't know you've done something wrong until an assertion on the C++ side crashes Unity.

Some confusion here I think. Your `CesiumGlobeFlightPath` does have a public default constructor. The C# compiler has generated it because you haven't explicitly declared any constructors at all. And because you're using it from `ConfigureReinterop.cs`, Reinterop has made it accessible from C++.

If you don't want any public constructors, you need to define a private one (even if it has no implementation). Then the line you added to ConfigureReinterop.cs will be a compiler error, and you'll want to simply remove it. Then I think everything will be working as expected, but let me know if not.

azrogers:
Thanks, Kevin! With that, I was able to make the constructor private and get everything working!

kring (commented):
Looks great to me, just a couple trivial comments.

azrogers:
@kring Resolved your comments!

native~/Runtime/src/CesiumSimplePlanarEllipsoidCurveImpl.h
```diff
@@ -0,0 +1,45 @@
...
+      const DotNet::Unity::Mathematics::double3 sourceLlh,
+      const DotNet::Unity::Mathematics::double3 destinationLlh);
+
+  DotNet::Unity::Mathematics::double3 GetPosition(
+      const DotNet::CesiumForUnity::CesiumSimplePlanarEllipsoidCurve& path,
+      double percentage,
+      double additionalHeight) const;
+
+private:
+  std::unique_ptr<CesiumGeospatial::SimplePlanarEllipsoidCurve> _curve;
```
kring, line 42: I noticed one more small thing worth fixing. It would be a bit better, I think, to make this a `std::optional` instead of a `std::unique_ptr`. The former avoids a heap allocation and just has nicer semantics overall (for example, it allows the `CesiumSimplePlanarEllipsoidCurveImpl` to be copyable; not that we're likely to copy it).

azrogers:
@kring Changed to `std::optional`!

native~/Runtime/src/CesiumSimplePlanarEllipsoidCurveImpl.cpp
```diff
@@ -25,12 +25,12 @@ bool CesiumSimplePlanarEllipsoidCurveImpl::
...
   if (!flightPath.has_value()) {
-    this->_curve = nullptr;
+    this->_curve = std::nullopt;
     return false;
   }
 
-  this->_curve = std::make_unique<SimplePlanarEllipsoidCurve>(*flightPath);
-  return this->_curve != nullptr;
+  this->_curve = *flightPath;
+  return this->_curve.has_value();
```
kring, line 33: This function (and the one below) can be simplified quite a bit now that you're using `std::optional`. Something like this (not tested):

```cpp
  this->_curve =
      SimplePlanarEllipsoidCurve::fromEarthCenteredEarthFixedCoordinates(
          Ellipsoid::WGS84,
          glm::dvec3(sourceEcef.x, sourceEcef.y, sourceEcef.z),
          glm::dvec3(destinationEcef.x, destinationEcef.y, destinationEcef.z));
  return this->_curve.has_value();
```

kring:
Looks great, just one last nitpick! :)

csciguy8:
Looks good @azrogers!.

Checked all remaining comments, retested, and a did a sanity check over all the file diffs once more.

Will merge once CI completes...

csciguy8 (approved):

PR #396, opened 2024-01-25
Allow specifying an activation distance for CesiumOriginShift instead of updating every frame

### LINKED ISSUE #182: Allow origin rebasing based on distance, rather than every frame
`CesiumOriginShift` currently shifts the origin every frame. We should add an option to only shift when the camera is X meters from the old origin.

azrogers:
Fixes #182.

Currently, `CesiumOriginShift` sets the origin of its parent georeference every frame. This change adds an activation distance property and a checkbox to enable it. If enabled, the origin will only be shifted when the position of the `CesiumGlobeAnchor` is greater than `activationDistance` meters away from the parent `CesiumGeoreference`. 

Originally, this change produced a strange bug with the `CharacterController` used by `CesiumCameraController`. I've fixed that bug by moving the movement code in `CesiumCameraController` to `FixedUpdate`, as it's manipulating a physics object. In addition, `Physics.SyncTransforms` needs to be called when updating the origin, so that the physics system is informed of the `CharacterController`'s new position.

I've also added a test for this bug as well as a test for the activation distance property itself.


Tests/TestCesiumOriginShift.cs
```diff
@@ -97,4 +97,96 @@ public IEnumerator UsesGeoreferenceTransformForActivationCheck()
...
+        originShift.useActivationDistance = true;
+        originShift.activationDistance = 5000;
+
+        yield return null;
+
+        globeAnchor.positionGlobeFixed = baseEcef + new double3(4999.0, 0, 0);
+
+        yield return null;
+
+        Assert.AreEqual(baseEcef.x + 4999.0, globeAnchor.positionGlobeFixed.x);
```
j9liu, line 126: Adding a small explanatory comment above here would be nice, like 
```
// The object is still within the distance threshold, so the georeference should remain unchanged.
```

Tests/TestCesiumOriginShift.cs
```diff
@@ -97,4 +97,96 @@ public IEnumerator UsesGeoreferenceTransformForActivationCheck()
...
+        Assert.AreEqual(baseEcef.x + 0, georeference.ecefX);
+
+        globeAnchor.positionGlobeFixed = baseEcef + new double3(5010.0, 0, 0);
+        yield return null;
+
+        Assert.AreEqual(baseEcef.x + 5010.0, globeAnchor.positionGlobeFixed.x);
```
j9liu, line 132: Adding a small explanatory comment above here would be nice, like 
```
// The object has surpassed the distance threshold, so the georeference should shift to the object's ECEF position.
```

Tests/TestCesiumOriginShift.cs
```diff
@@ -97,4 +97,96 @@ public IEnumerator UsesGeoreferenceTransformForActivationCheck()
...
+        Assert.AreEqual(baseEcef.x + 5010.0, georeference.ecefX);
+    }
+
+    // Testing a bug where the character controller would cause a jump ahead when the activation distance is hit.
+    [UnityTest]
+    public IEnumerator ShiftingOriginWithCharacterController()
```
j9liu, line 138: Nitpick for naming consistency with other unit tests 
```suggestion
    public IEnumerator ShiftsOriginWithCharacterController()
```

Tests/TestCesiumOriginShift.cs
```diff
@@ -97,4 +97,96 @@ public IEnumerator UsesGeoreferenceTransformForActivationCheck()
...
+            controller.Move((float3)movement);
+
+            yield return new WaitForEndOfFrame();
+
+            globeAnchor.Sync();
+            Assert.AreEqual((float)(previousPositionEcef.x + unitsEcef), (float)globeAnchor.positionGlobeFixed.x);
+            Assert.Less(georeference.ecefX - globeAnchor.positionGlobeFixed.x, 5000.0);
+
+            Debug.Log($"georeference d: ({georeference.ecefX - baseEcef.x}, {georeference.ecefY - baseEcef.y}, {georeference.ecefZ - baseEcef.z})");
+            Debug.Log($"anchor d: {globeAnchor.positionGlobeFixed - baseEcef}");
```
j9liu, line 189: Remove?

Tests/TestCesiumOriginShift.cs
```diff
@@ -97,4 +97,96 @@ public IEnumerator UsesGeoreferenceTransformForActivationCheck()
...
+            yield return new WaitForFixedUpdate();
+
+            double unitsEcef = speed * Time.deltaTime;
+            double3 movement = georeference.TransformEarthCenteredEarthFixedDirectionToUnity(new double3(unitsEcef, 0, 0));
+            controller.Move((float3)movement);
+
+            yield return new WaitForEndOfFrame();
+
+            globeAnchor.Sync();
+            Assert.AreEqual((float)(previousPositionEcef.x + unitsEcef), (float)globeAnchor.positionGlobeFixed.x);
```
j9liu, line 185: For these float-arithmetic checks, can we use epsilon equality just to be safe? Similar to this test in `TestCesiumGeoreference`

https://github.com/CesiumGS/cesium-unity/blob/db4df38750eae7e5ed8be4b9669d2171c8345f80/Tests/TestCesiumGeoreference.cs#L71

Tests/TestCesiumOriginShift.cs
```diff
@@ -97,4 +97,96 @@ public IEnumerator UsesGeoreferenceTransformForActivationCheck()
...
+        originShift.activationDistance = 5000;
+
+        yield return null;
+
+        globeAnchor.positionGlobeFixed = baseEcef + new double3(4999.0, 0, 0);
+
+        yield return null;
+
+        Assert.AreEqual(baseEcef.x + 4999.0, globeAnchor.positionGlobeFixed.x);
+        Assert.AreEqual(baseEcef.x + 0, georeference.ecefX);
```
j9liu, line 127: Nitpick but we can just simplify this
```suggestion
        Assert.AreEqual(baseEcef.x, georeference.ecefX);
```

Tests/TestCesiumOriginShift.cs
```diff
@@ -97,4 +97,96 @@ public IEnumerator UsesGeoreferenceTransformForActivationCheck()
...
+
+        globeAnchor.positionGlobeFixed = baseEcef + new double3(5010.0, 0, 0);
+        yield return null;
+
+        Assert.AreEqual(baseEcef.x + 5010.0, globeAnchor.positionGlobeFixed.x);
+        Assert.AreEqual(baseEcef.x + 5010.0, georeference.ecefX);
```
j9liu, line 133: Nitpick, but I think it makes sense to check between the globe anchor + georeference at this point

```suggestion
        Assert.AreEqual(globeAnchor.positionGlobeFixed.x, georeference.ecefX);
```

j9liu (commented):
Thanks @azrogers ! I left some comments; let me know if you have any questions or disagree with any of them.

Also I'm noticing that the framerate seems to drop more often on this branch than on main, when I'm just moving around with the DynamicCamera. But I can't confirm for sure. Have you noticed this at all?

Runtime/CesiumOriginShift.cs
```diff
@@ -42,13 +69,15 @@ void LateUpdate()
...
+            this.UpdateFromEcef(georeference, anchor);
         }
 
         private List<CesiumSubScene> _sublevelsScratch = new List<CesiumSubScene>();
 
-        private void UpdateFromEcef(CesiumGeoreference georeference, double3 ecef)
+        private void UpdateFromEcef(CesiumGeoreference georeference, CesiumGlobeAnchor anchor)
         {
+            double3 ecef = anchor.positionGlobeFixed;
+
```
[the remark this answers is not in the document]
azrogers: I thought I needed to access the anchor while trying things out, but it looks like in the final version I didn't end up needing it. I'll change it back.

azrogers:
@j9liu Updated based on review! I tried comparing profiler results from before and after the changes but nothing seems to jump out at me. I can't tell myself if it's slower or not.

j9liu:
Thanks @azrogers !

PR #398, opened 2024-01-30
Add flag to TextureLoader for sRGB textures, load some textures as linear

### LINKED ISSUE #332: All textures are created as sRGB, when some need to be linear
`TextureLoader.cpp` passes `false` for the `linear` parameter, so it always created sRGB textures. This is correct for most but not all of the textures we'll be creating. In Cesium for Unreal, a flag is passed into the texture creation function to indicate whether or not it should be sRGB, and the same needs to happen here in Cesium for Unity.

azrogers:
Fixes #332.

Currently, all textures in Unity are loaded as if they're sRGB textures. This isn't correct for things like normal maps, where treating them as sRGB will mess up the direction of the normals. This change adds a flag to `TextureLoader::loadTexture` to allow specifying which textures are sRGB, and specifies the flag for each texture loaded based on the Unreal implementation.


kring:
Looks great! Please just merge in main. It's another painful conflict, unfortunately.

azrogers:
@kring Merged with main! It feels a little bit like I did something wrong in the merge but the changes look right to me, so 🤷‍♀️

j9liu:
Thanks @azrogers !

PR #399, opened 2024-01-30
Use Delayed input fields for `CesiumGeoreference` editor

### LINKED ISSUE #324: Reduce CesiumGeoreference updates when manually entering coordinates
Install Unity Hub 3.4.2 and Unity 2021.3.25f1.

Follow the 'Cesium for Unity Quickstart" guide,
https://cesium.com/learn/unity/unity-quickstart/

1. When creating the new project, use the "3D Sample Scene (HDRP)" template.
2. When installing Cesium from the Package Manager, use "Cesium for Unity, Version 1.2.0 - May 09, 2023"
3. Continue with the tutorial until changing the CesiumGeoreference coordinates in **Step 5: Add global 3D buildings to your scene**
4. Select your CesiumGeoreference game object in the Hierarchy
5. In the Inspector Window, find "Latitude", select the existing value, press delete
6. Enter the new value, 41.878101, one digit at a time

Notice that the world geometry updates on every key event, from deleting the existing value and on every keystroke. 

It might be a nicer experience if the terrain only updated when you press enter, or click on another window (change focus).

azrogers:
Fixes #324.

This change uses `EditorGUI.DelayedDoubleField` and the `Delayed` attribute to reduce the number of updates when editing the properties of a `CesiumGeoreference` in the inspector.

### DIFF
--- a/Editor/CesiumGeoreferenceEditor.cs
+++ b/Editor/CesiumGeoreferenceEditor.cs
@@ -140,7 +140,8 @@ namespace CesiumForUnity
                 this._latitude,
                 -90.0,
                 90.0,
-                latitudeContent);
+                latitudeContent,
+                delayed: true);
 
             GUIContent longitudeContent = new GUIContent(
                 "Longitude",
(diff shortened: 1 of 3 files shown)


j9liu (commented):
Works great @azrogers . Thanks!

PR #400, opened 2024-02-01
Better activation radius gizmo for CesiumSubScene

### LINKED ISSUE #83: Visualize CesiumSubScene activation radius
In Cesium for Unreal, there's an "Enable Load Radii" option on the CesiumGeoreference. When enabled, sub-level radii are visualized in the Editor as spheres.

We should add a similar "Show Activation Radius" option to `CesiumSubScene` in Cesium for Unity.

azrogers:
Closes #83.

`CesiumSubScene` currently uses `Gizmos.DrawWireSphere` to visualize the activation radius of the subscene. This doesn't help very much, since `Gizmos.DrawWireSphere` really only draws three wire circles representing the sphere's radii, and that's it. The simplest way to do better is to grab the built-in sphere mesh (the one that gets used when you create a sphere using `GameObject > 3D Object > Sphere`) and draw that in wireframe.

### DIFF
--- a/Runtime/CesiumSubScene.cs
+++ b/Runtime/CesiumSubScene.cs
@@ -21,6 +21,11 @@ namespace CesiumForUnity
     [IconAttribute("Packages/com.cesium.unity/Editor/Resources/Cesium-24x24.png")]
     public class CesiumSubScene : MonoBehaviour
     {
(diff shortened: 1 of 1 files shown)


j9liu (commented):
Thanks @azrogers !

PR #402, opened 2024-02-02
Fix UnityPrepareRendererResources trying to destroy texture assets

### LINKED ISSUE #344: Using copies of the default materials results in error log spam
Multiple users have tried creating their own copies of our shaders for their projects, only to get this error spammed in the console.

image
>"Destroying assets is not permitted to avoid data loss."

Using `Resources.Load` gives us an object that is linked to the asset itself, so we don't want to destroy a material returned by that function. But we always `Instantiate` that material, so one would think the material is safe to destroy.

It's also hard to reliably reproduce this bug. I remember getting it to show up at one point on Unity 2021.3.14f, but I don't have concrete steps. I also haven't been able to reproduce it on Unity 2021.3.26f.

A user commented that they were able to get rid of the bug by "deleting the reference to “green1x1.png” in the “metalicRoughnessTexture”". So maybe we're incorrectly trying to destroy the texture reference?

Known forum posts: [X](https://community.cesium.com/t/how-to-limit-tiles-within-a-certain-radius/25137/6), [X](https://community.cesium.com/t/error-message-for-modifying-shader-destroying-assets-is-not-permitted-to-avoid-data-loss/23281)

azrogers:
Fixes #344.

Currently, when UnityPrepareRendererResources frees a tile, it calls `DestroyImmediate` on the material and all of its textures. This is fine and in fact desired behavior for the materials and textures Cesium creates at runtime - we need to make sure these texture resources get cleaned up so we don't leak memory. This is also fine for user-provided materials, as we create a new instance of the material - as long as the material doesn't have any texture assets set for its texture properties. Textures aren't copied when you copy a material, so when `freePrimitiveGameObject` goes through all of the material's texture properties and calls `Destroy` on everything it finds, it ends up attempting to destroy the assets the end user specified on the material. This produces a "Destroying assets is not permitted to avoid data loss" error.

There are two ways we can solve this. The first way would be to make a new instance of all the textures the same way we make a new instance of the materials. However, it would be simpler to just keep track of all the textures we're creating at runtime, and just not try to destroy textures we didn't create. There's a few ways to do this, but I went with the simplest - setting the `UnityEngine::HideFlags::HideAndDontSave` hideFlag on each texture created, and checking it before destroying. This prevents assets from being destroyed.

One caveat is that if the user sets a texture *they* created on the material that also has `UnityEngine::HideFlags::HideAndDontSave` set, our code will attempt to destroy it too. This won't result in errors, just undesirable behavior. However, I think this possibility is pretty slim, and in the case where it's an issue we can suggest that they simply use `UnityEngine::HideFlags::DontSave` instead.

### DIFF
--- a/native~/Runtime/src/UnityPrepareRendererResources.cpp
+++ b/native~/Runtime/src/UnityPrepareRendererResources.cpp
@@ -1191,7 +1191,8 @@ void* UnityPrepareRendererResources::prepareInMainThread(
               if (texCoordIndexIt != primitiveInfo.uvIndexMap.end()) {
                 UnityEngine::Texture texture =
                     TextureLoader::loadTexture(gltf, baseColorTexture->index);
-                texture.hideFlags(DotNet::UnityEngine::HideFlags::DontSave);
+                texture.hideFlags(
+                    DotNet::UnityEngine::HideFlags::HideAndDontSave);
                 if (texture != nullptr) {
                   material.SetTexture(
                       shaderProperty.getBaseColorTextureID(),
@@ -1211,7 +1212,8 @@ void* UnityPrepareRendererResources::prepareInMainThread(
               if (texCoordIndexIt != primitiveInfo.uvIndexMap.end()) {
                 UnityEngine::Texture texture =
                     TextureLoader::loadTexture(gltf, metallicRoughness->index);
-                texture.hideFlags(DotNet::UnityEngine::HideFlags::DontSave);
+                texture.hideFlags(
+                    DotNet::UnityEngine::HideFlags::HideAndDontSave);
                 if (texture != nullptr) {
                   material.SetTexture(
                       shaderProperty.getMetallicRoughnessTextureID(),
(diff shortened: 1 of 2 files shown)


native~/Runtime/src/UnityPrepareRendererResources.cpp
```diff
@@ -1191,6 +1191,8 @@ void* UnityPrepareRendererResources::prepareInMainThread(
               if (texCoordIndexIt != primitiveInfo.uvIndexMap.end()) {
                 UnityEngine::Texture texture =
                     TextureLoader::loadTexture(gltf, baseColorTexture->index);
+                texture.hideFlags(
+                    DotNet::UnityEngine::HideFlags::HideAndDontSave);
```
kring, line 1195: This needs to be inside the `if` below, because it's possible for `loadTexture` to return a null reference.

kring (commented):
This looks good to me. Just one comment below. Could you also please merge in main? Unfortunately you'll probably have to reapply the changes after #385 was merged. Sorry about that!

azrogers:
@kring Merged with main and addressed your comment!

kring:
Thanks @azrogers! Lots of people will be happy to see this one fixed!

PR #404, opened 2024-02-07
Add `materialKey` to `CesiumRasterOverlay`

### LINKED ISSUE #401: Support reordering raster overlays
Raster overlays are currently applied in the order in which their OnEnable is called. There should be more explicit control over their order. For example, it should be possible to enable a layer at runtime that is drawn in between two existing ones.

j9liu:
Depends on #385 so merge that first.

Fixes #401.

This PR adds the `materialKey` property to `CesiumRasterOverlay` and intends to be analogous to material layer keys in Cesium for Unreal. In the tileset materials, overlay parameters now have the following syntax:

- `_overlayTexture_KEY`
- `_overlayTextureCoordinateIndex_KEY`
- `_overlayTranslationAndScale_KEY`

Where `KEY` is the string in `materialKey`. On the native side, raster overlays store `materialKey` in their name so that they can be attached / detached from the correct variables.

The UI for `CesiumRasterOverlay` has been updated to reflect this property. However, instead of the typical text field, it displays a popup that reflects which keys are actually present in the material.

Finally, I also took the time to add this little warning in the `CesiumRasterOverlay`. It appears if the raster overlay is not attached to a `Cesium3DTileset`. (might be nice to use this in other components too!)


kring:
This looks great! One small problem I noticed: it looks like this PR is accidentally undoing a cesium-native upgrade.

kring:
Thanks @j9liu!

PR #405, opened 2024-02-08
Support textures with less than four channels

j9liu:
Depends on #404, so merge that first.

While working on polygon clipping, I saw that the rasterized polygon textures were importing incorrectly, appearing very corrupted. In `TextureLoader.cpp`, we were not supporting texture formats with less than four pixels, which was causing problems for the single-channel `RasterPolygonOverlay` textures.

This PR adds support for it. I did this on my `polygon-clipping` branch, but didn't want this change to be dependent on it.


kring:
👍 I'll merge this one after #404.

Issue #64, opened 2022-10-14
Add support for RasterizedPolygonsTileExcluder


mwilsnd:
Currently tile exclusion is not exposed to C#. Support for RasterizedPolygonsOverlay/RasterizedPolygonsTileExcluder enables extra optimizations for developers creating more confined game levels or AR experiences. Going the extra mile, support for C# implementations of ITileExcluder and the ability to provide these to TilesetOptions::excluders would provide maximum flexibility for custom culling logic without having to leave the C# environment.

j9liu:
For the record, support for `ITileExcluder` implementations was added with #248, but we still don't support `RasterizedPolygonsTileExcluder` itself.

XinDongDeXin:
Has this problem been solved now?

XinDongDeXin:
Maps load until the system crashes, and it would be nice if there was a way to limit the load range.

kring:
j9liu's comment above still represents the current state. The feature described in #248 should work pretty well for the use-case of limiting range.

But if normal map loading is leading to a system crash (out of memory?), then something is almost certainly wrong either with your settings or your tileset. Perhaps you've set the Maximum Screen Space Error too low? Or your tileset has too many tiles at even a low level of detail? Or the Maximum Cache Bytes is set too high for the available system memory? Keep in mind that the value represents the size of the tiles coming over the network; they can actually be significantly larger once resident in memory. Set it to 0 and see if the crash goes away.

j9liu:
This should have been closed by #407 , closing now.

PR #407, opened 2024-02-09
Add support for polygon clipping

### LINKED ISSUE #360: UE “Cesium Cartographic Polygon”
Add UE “Cesium Cartographic Polygon” to unity

j9liu:
Fixes #360 #64. Depends on #405 (which itself depends on material keys), so merge that first.

This PR adds support for rasterized polygon clipping. 

**DISCLAIMER**: This requires the Splines package, which is only available in Unity 2022.2 or later. It's not listed as a package dependency, so our package will still load in 2021.3. However, this functionality will be inaccessible in the older versions.

## Workflow

The workflow is similar to Cesium for Unreal's. Users can add a `CesiumCartographicPolygon` from the quick add panel.

This will spawn a game object with a `CesiumCartographicPolygon` component, which automatically adds a `Spline` and `CesiumGlobeAnchor` component. You can modify the game object's transform, add or remove points from the spline component, and edit the spline itself by clicking this icon in the Scene View.

image

Note that the spline is required to be closed, and completely linear. If either of these conditions is not true, the polygon will log errors. Interestingly, the Spline component supports multiple splines. For simplicity, only the first one is rasterized.

Then, you can connect this polygon to a tileset using the `CesiumPolygonRasterOverlay` component. The settings are exactly the same as in Unreal.

The materials have been modified to handle alpha clipping via "Clipping" overlay parameters. The final result:

| Normal | Inverted Selection |
| -------- | -------------------- |
| image | image |

## Splines package missing (Unity 2022.2+) 

The assembly definition has been modified to reference the `com.unity.splines` package. If it isn't detected, then the `SUPPORTS_SPLINES` won't be defined. Users will still be able to add `CesiumCartographicPolygon` and `CesiumPolygonRasterOverlay` through the UI, but they will appear with these warnings:

## Splines package unavailable (Unity 2021.3f)

`com.unity.splines` is not listed `package.json` as a dependency, so Cesium for Unity isn't blocked from loading in earlier versions. When it does load, there will be no UI to add `CesiumCartographicPolygon`, and `CesiumPolygonRasterOverlay` is hidden from the component menu. If somehow they make their way into the scene, these warnings will appear:

image


Runtime/CesiumPolygonRasterOverlay.cs
```diff
@@ -0,0 +1,109 @@
...
+            {
+                this._excludeSelectedTiles = value;
+                this.Refresh();
+            }
+        }
+
+#if UNITY_EDITOR
+        private void Reset()
+        {
+            this.materialKey = "Clipping";
```
j9liu, line 100: Annoyingly, when you first add the overlay, its material key will be initialized with the default value in `CesiumRasterOverlay`. `Reset` happens after `OnEnable`, so it will potentially log a warning that it shares a material key with another overlay on the tileset.

I'll see if I can try doing something similar to `CesiumSubScene` for circumventing this.

kring:
Also, the created Spline Container in the above scenario had default settings, which made it hard to use.

kring:
This looks really good! One thing I want to test is that the packaged version of the plugin works just as well, in both 2021.3 and 2022.2. But to do that, we need a working build from CI, which is proving elusive. At the moment, the problem is that Chocolatey is down, so none of our self-hosted builds can start. 😞

j9liu:
Thanks for the review @kring ! I just pushed some code that should improve the case where a polygon is created before the Splines package is added.

kring:
>  I just pushed some code that should improve the case where a polygon is created before the Splines package is added.

I tried this out and it looks good!

I also tested the release package in 2021.3 and 2022.3, and it also works well in both cases.

It's pretty amazing how well this turned out, and how quickly. Nice job!

PR #413, opened 2024-02-15
Fix CesiumOriginShift test currently breaking builds

azrogers:
Currently the `ShiftsOriginWithCharacterController` test for `CesiumOriginShift` is failing, causing builds to fail. Currently, the test's results depend heavily on frame time - it uses `Time.deltaTime` to calculate movement distance, and the rest runs for a specific number of seconds, rather than a specific number of frames. This means that a machine that can render 30 frames per second will get a different test result than a machine that can only render 15. I've changed the test to execute a set number of frames and to use `Time.fixedDeltaTime` for its movement, so the test should produce a same result no matter the frame time.

### DIFF
--- a/Tests/TestCesiumOriginShift.cs
+++ b/Tests/TestCesiumOriginShift.cs
@@ -164,24 +164,23 @@ public class TestCesiumOriginShift
 
         yield return new WaitForEndOfFrame();
 
-        IEqualityComparer<double> epsilon6 = Comparers.Double(1e-6, 1e-6);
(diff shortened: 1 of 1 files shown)


azrogers:
For some reason it's taking 180 seconds now? I'll look into that...

kring:
I've changed the test to make it run much more quickly, and hopefully be more reliable. It's best to avoid tests that take 10 seconds if we can help it, because test time really adds up as we add more tests. I'm going to optimistically merge this in the hope that it will make CI green in main!

PR #420, opened 2024-02-20
Switch to GitHub-hosted runners

kring:
Instead of our self-hosted runners. In addition to saving money, this should give us more consistent build times.

This also includes #410.


kring:
I'm going to merge this immediately in order to try go get green Unity builds again.

PR #422, opened 2024-02-20
Change `CesiumCameraController` and `CesiumFlyToController` to use `CesiumGlobeAnchor` on parent GameObjects

### LINKED ISSUE #415: Allow `CesiumGlobeAnchor` to be attached to parent object in `CesiumCameraController`
`CesiumCameraController` currently expects the GameObject it's attached to to also have a `CesiumGlobeAnchor` component. It then creates a `CharacterController` on this GameObject and uses that controller to move the camera around, with the required `CesiumOriginShift` keeping the `CesiumGeoreference` informed. 

Now, here's the problem. I have a hierarchy that looks like this:
```
- CesiumGeoreference
    - CesiumGlobeAnchor, CesiumOriginShift
        - XROrigin
            - Main Camera
```

`XROrigin` *must* be above the camera, as it's the reference point for (0, 0, 0) in the real world. For example, if you start up the app in your chair and then walk across the room, the `XROrigin` remains at your chair but the `Main Camera` has translated and rotated along with your head as you moved. Also, the `XROrigin` *must* be georeferenced, to maintain its attachment to the real-world space it's representing. Otherwise, if you sit in that same chair and fly across the earth, you'll end up with the origin hundreds of thousands of units away, breaking that spatial connection and running into precision errors to boot. You might see the problem here - we can't have the `CesiumGlobeAnchor` and `CesiumCameraController` on the same object.

The `CesiumCameraController` should treat the parent `CesiumGlobeAnchor` as the anchor to modify and create the character controller on, rather than assuming it will always be on the same GameObject. This could be accomplished by using `GetComponentInParent` instead of `GetComponent` for `this._globeAnchor`, and creating the `CharacterController` on the same GameObject as the `this._globeAnchor`.

azrogers:
Closes #415.

Currently, the `CesiumCameraController` and `CesiumFlyToController` expect to be on an object with a `CesiumOriginShift` and `CesiumGlobeAnchor` present. This makes it difficult to setup on XR rigs, as detailed in the issue above. This changes both classes to use `CesiumGlobeAnchor` components present on parent objects, while using the `Reset` event to still create an anchor if none is present, as was the previous behavior.


csciguy8 (approved):
Looks good @azrogers!

Locally tested unity samples and ran the unit tests. Code looks to preserve the original behavior like you stated.

Will merge once CI completes...

PR #423, opened 2024-02-20
Fix test failure by yielding null instead of WaitForEndOfFrame.

kring:
This fixes the test failure on CI and on my own machine when running the tests from the command-line the way that CI does. I can't entirely explain why `WaitForEndOfFrame` causes problems, though.


kring:
All green, so I'm merging.

PR #425, opened 2024-02-23
Support custom input actions on `CesiumCameraController`

### LINKED ISSUE #414: Support for arbitrary input methods for `CesiumCameraController`
Currently, `CesiumCameraController` is set up to use a hardcoded `InputActionMap` that provides bindings for keyboard, mouse, and gamepad. If a user wants to support a new input method - for example, an XR controller - they have to copy the class and make their own modifications. 

I think we should try to find a way to support arbitrary input methods. While `CesiumCameraController` isn't a "core" part of Cesium's technology by any means, it is one of the first things people interact with when starting a new project with Cesium for Unity, and making it easier to integrate new input methods would make it easier for people to make new AR and VR projects with Cesium.

I can think of two ways we can go about this:

1. Using Unity's Input System

We're currently using Unity's Input System package in `CesiumCameraController`, falling back to the old `UnityEngine.Input` class when it's not available. Though we currently hardcode an `InputActionMap` for use with the camera controller, we could simply expose this as a property to the inspector, so that a user who wants to support a new controller can simply make a new Input Actions asset that defines the mapping.

This would be the simplest way to support new inputs, with only a few code changes needed. The disadvantage of this approach is that it makes it more difficult to do anything more complex than simply read a value from an input source and pass it on to the program. For example, specifying that you want the camera to move in the direction that the controller is pointed rather than the forward vector would require [writing a custom input processor](https://docs.unity3d.com/Packages/com.unity.inputsystem@1.0/manual/Processors.html#writing-custom-processors). But this is doable.

2. Using an Additional Component

The other way to support new input methods would be to entirely detach input from `CesiumCameraController` using an interface and an additional component. This interface would have methods for things like `GetMovementVector()`, `GetSpeedChange()`, etc. We would provide a component out of the box implementing this interface that would use the hardcoded mapping we currently have, and then new input methods could be specified by writing a new implementation of the interface.

The advantage of this approach is that it would support pretty much every input method and scenario imaginable. Things like, for example, performing a raycast to get the movement vector, would be a lot easier to implement in this method than using an `InputActionMap`. This, of course, comes at the cost of accessibility to the end user.

azrogers:
Fixes #414.

This change adds a set of properties to `CesiumCameraController` that allows specifying an input action to bind to each action in the camera controller (such as look, move, speed change, etc). This property can specifying a binding for that action, or reference an existing action to use for the input. If no action is specified, `CesiumCameraController` will create default bindings for that action as it did before this change.

### DIFF
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -20,6 +20,7 @@
 - Added a `distance` property to `CesiumOriginShift`, which specifies the maximum allowed distance from the current origin before it is shifted.
 - Added support for the `KHR_texture_transform` glTF extension - including rotation - in `baseColorTexture`, `metallicRoughnessTexture`, `emissiveTexture`, `normalTexture`, and `occlusionTexture`. The transformation is now applied on the GPU via nodes in the Material, rather than on the CPU by directly modifying texture coordinates.
 - Added `materialKey` to `CesiumRasterOverlay`, which matches the overlay to its corresponding parameters in the tileset's material. This allows for explicit ordering of raster overlays and overlay-specific effects. 
+- `CesiumCameraController` can now accept custom input actions that override the default inputs.
 
 ##### Fixes :wrench:
 
--- a/Editor/CesiumCameraControllerEditor.cs
+++ b/Editor/CesiumCameraControllerEditor.cs
@@ -119,7 +119,7 @@ namespace CesiumForUnity
             GUILayout.EndHorizontal();
 
             EditorGUI.EndDisabledGroup();
-            
+
             EditorGUILayout.Space(5);
 
             GUILayout.BeginHorizontal();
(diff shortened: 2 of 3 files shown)


azrogers:
Updated CHANGES.md. In response to your two questions:

1. `InputActionProperty` allows the same UI as `InputAction`, with the addition of allowing `InputActionReference` assets to be specified in place of a new binding. So I think the name "Input Actions" is still accurate, it will work the same as an `InputAction` if you want to use it that way.
2. I don't think the `InputActionProperty` fields can be set with field initializers. Could possibly be set through the `Reset` method in the editor? But I don't think `Reset` runs on existing components, only when you add a new component to the scene, so any camera controllers in scenes people already made before this change would stop working. So we'd need some way of setting these default values on components already added to scenes too.

PR #429, opened 2024-03-05
Use rotation from globe anchor in CesiumFlyToController

azrogers:
Something that got overlooked in changing CesiumFlyToController to be able to use a parent CesiumGlobeAnchor. Instead of using the local transform, we want to use the rotation of whatever transform the CesiumGlobeAnchor is attached to for the source rotation on our flight, since this is the object that the rotation is going to be applied to as it flies.


j9liu:
Sorry for the very late, completely out-of-context post, but what was the reason for the change to use a parent? Is there a reason you can't use a `CesiumFlyToController` on a parent object? In other words, why would a parent be anchored, whereas a child object would be the one flying?

j9liu:
Ignore the above. We talked about this offline and I opened https://github.com/CesiumGS/cesium-unity/issues/434 for it.

PR #445, opened 2024-03-22
Don't delete `CesiumCreditSystem` while unloading other scenes

### LINKED ISSUE #436: CesiumCreditSystem destroys itself when an additive scene is unloaded
My main scene uses Cesium to show a map and occasionally loads additional Unity scenes (that have nothing to do with the Cesium content) in additive mode. When one of those unrelated additive scenes is unloaded the CesiumCreditSystem destroys itself, leaving my scene without credits.

j9liu:
Fixes #436. This PR prevents `CesiumCreditSystem` from being destroyed in its scene when other additive scenes are loaded / unloaded.

To test, I created a new scene called "TestScene" and put a cube in it. I added it under the scenes in Build Settings, then created a script called "LoadUnloadScene.cs" with this code:

```
using UnityEngine;
using UnityEngine.InputSystem;
using UnityEngine.SceneManagement;

public class LoadUnloadScene : MonoBehaviour
{
    void Start()
    {
        SceneManager.LoadScene("TestScene", LoadSceneMode.Additive);
    }

    // Update is called once per frame
    void Update()
    {
        if (Keyboard.current.digit1Key.wasPressedThisFrame)
        {
            SceneManager.UnloadSceneAsync("TestScene");
        }
    }
}
```

Then, I attached `LoadUnloadScene` as a component to the `CesiumGeoreference`. When you first run in Play mode, the test scene will load in. After pressing "1", the scene will unload. The credits UI should not disappear.

I even got this to work when the additive scene contained a tileset! In the first Cesium for Unity Sample scene, I moved Cesium OSM Buildings to a new georeference (same coordinates) under TestScene. Loading / unloading the scene properly updates the credit system in the original scene, without removing the UI. You can enable "Show Credits On Screen" to confirm that the original credit system updates for the tileset in the additive scene.

One problem here is, this all assumes the original credit system will not be unloaded. It's possible (though I don't know how practical) that the original scene is unloaded while additive ones remain. I tried fixing this but can't get it to work quite yet, so I'll write an issue and naybe address it in a follow-up PR.

### DIFF
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -5,6 +5,7 @@
 ##### Fixes :wrench:
 
 - Normal, metallic-roughness, and occlusion textures from glTF models will now be correctly treated as linear rather than sRGB.
+- Fixed a bug where `CesiumCreditSystem` would delete itself from its scene when other additive scenes were unloaded.
 
 ### v1.8.0 - 2023-03-01
 
--- a/Runtime/CesiumCreditSystem.cs
+++ b/Runtime/CesiumCreditSystem.cs
@@ -7,7 +7,6 @@ using UnityEngine.SceneManagement;
 using UnityEngine.Networking;
 
 #if UNITY_EDITOR
-using UnityEditor;
 using UnityEditor.SceneManagement;
 #endif
 
--- a/Runtime/CesiumCreditSystem.cs
+++ b/Runtime/CesiumCreditSystem.cs
@@ -105,6 +105,7 @@ namespace CesiumForUnity
         const string defaultName = "CesiumCreditSystemDefault";
         const string creditSystemPrefabName = "CesiumCreditSystem";
 
+        #region Fields and Events
         /// <summary>
         /// The current on-screen credits, represented as <see cref="CesiumCredit"/>s.
         /// </summary>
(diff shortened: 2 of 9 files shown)


j9liu:
To address the case where the original scene is unloaded, I pushed a commit that recreates the credit system if it's no longer valid on a tileset. Similar to Unreal, the tileset will recreate itself with respect to the new credit sytsem.

I _thought_ the UI was bugged when the scene unloaded. But then I remembered that Cesium only works with the camera tagged "MainCamera". So I was deleting the MainCamera in the original scene, without assigning the tag to the new one. When you assign the tag to a new camera, the credits UI will reappear.

kring:
I took a look at the code and it all looks good to me. I haven't actually tried building/running/playing with it yet, though. I'll do that on Monday if Ashley doesn't beat me to it.

azrogers:
Thanks @j9liu! The changes look good to me, and seem to work correctly in my testing in the editor.

j9liu:
Thanks @azrogers! I noticed that the changelog hadn't been updated after the release, so I just pushed a commit to main to correct it.

PR #446, opened 2024-03-26
Add extra WMTS properties

j9liu:
This PR follows #430, which added basic support for WMTS overlays. The goal of this PR was to achieve parity with Cesium for Unreal, which boasts many more settings on the WMTS overlay.

Here's a screenshot of the new UI:

There are some differences from the Unreal one:
- Instead of the `useWebMercatorProjection` or `useGeographicProjection` flag, I made it formal enum so that the intent was much clearer.
- I've prefixed the "West", "South", etc. variables so they now read "Rectangle West", "Rectangle South", etc. I'm open to better names, but I thought it was better to be explicit in this way.

I'd like to update the Unreal UI to match Unity's, namely the enum change, but it's not an urgent change that has to make it into the next release.

### DIFF
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -2,6 +2,10 @@
 
 ### ? - ?
 
+##### Additions :tada:
+
+- Added `CesiumWebMapTileServiceRasterOverlay`, which enables Web Map Tile Service (WMTS) imagery to be draped on a `Cesium3DTileset`.
+
 ##### Fixes :wrench:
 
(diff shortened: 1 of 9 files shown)


j9liu:
@tal952 since you originally implemented #430, let me know if you disagree with any of the additions or changes I've made here!

kring:
Code and UI look great, and I tested it successfully with the "USGS Shaded Relief" WMTS in the "Imagery Layers Manipulation" CesiumJS Sandcastle example.

PR #449, opened 2024-04-02
Support in Unity for Cesium ion servers running in single-user authentication mode.

azrogers:
Requires https://github.com/CesiumGS/cesium-native/pull/841.

As described in the above ticket, we don't currently support Cesium ion Self-Hosted servers that are running in single-user authentication mode. This is inconvenient, especially since single-user mode is both the easiest to setup and the default authentication mode in Cesium ion Self-Hosted.

This PR changes the behavior of Cesium for Unity to request the current authentication mode from the /appData endpoint before connecting, ignoring the authentication flow and token options if in single-user mode.


j9liu:
@azrogers I get a crash running this branch when trying to resume an ion session. I believe it's because of `CesiumIonSessionImpl::Resume` at line 304: this blog of code:

https://github.com/CesiumGS/cesium-unity/blob/9be3acb7ef6cee0665555c2b4913d9beb5a0c770/native~/Editor/src/CesiumIonSessionImpl.cpp#L304-L310

`this->_appData` is nullopt, which is causing an accessor error.

native~/Editor/src/CesiumIonSessionImpl.cpp
```diff
@@ -254,6 +306,7 @@ void CesiumIonSessionImpl::Resume(
           this->_asyncSystem,
           this->_pAssetAccessor,
           userAccessToken.ToStlString(),
+          this->_appData.value(),
```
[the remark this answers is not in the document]
j9liu: Maybe something like this:

```
 CesiumAsync::Future<
      CesiumIonClient::Response<CesiumIonClient::ApplicationData>>
      futureAppData = CesiumIonClient::Connection::appData(
          this->_asyncSystem,
          this->_pAssetAccessor,
          server.apiUrl().ToStlString());

  std::move(futureAppData)
      .thenInMainThread(
          [userAccessToken, server, session, this](
              CesiumIonClient::Response<CesiumIonClient::ApplicationData>&&
                  appData) {
            CesiumAsync::Promise<void> promise =
                this->_asyncSystem.createPromise<void>();

            this->_appData = appData.value;
            if (!appData.value.has_value()) {
              promise.reject(std::runtime_error(fmt::format(
                  "Failed to obtain ion server application data: {}",
                  appData.errorMessage)));
            } else {
              promise.resolve();
            }

            return promise.getFuture();
          })
      .catchInMainThread([this](std::exception&& e) {
        logResponseErrors(e);
        this->_isResuming = false;
      })
      .thenInMainThread([this, userAccessToken, server]() {
        return this->_asyncSystem
            .createResolvedFuture<CesiumIonClient::Connection>(
                CesiumIonClient::Connection(
                    this->_asyncSystem,
                    this->_pAssetAccessor,
                    userAccessToken.ToStlString(),
                    this->_appData.value(),
                    server.apiUrl().ToStlString()));
      })
```

native~/Editor/src/CesiumIonSessionImpl.cpp
```diff
@@ -657,6 +711,13 @@ const std::vector<CesiumIonClient::Token>& CesiumIonSessionImpl::getTokens() {
   }
 }
 
+const CesiumIonClient::ApplicationData& CesiumIonSessionImpl::getAppData() {
+  if (this->_appData) {
+    return *this->_appData;
+  }
+  return CesiumIonClient::ApplicationData();
```
j9liu, line 718: You can conveniently simplify this with `value_or`!

```suggestion
  return this->_appData.value_or(CesiumIonClient::ApplicationData());
```

Editor/CesiumEditorWindow.cs
```diff
@@ -146,12 +148,16 @@ void DrawCesiumToolbar()
                 "Upload a tileset to Cesium ion to process it for efficient streaming to Cesium for Unity",
                 true,
                 this.UploadToIon);
+
+            EditorGUI.BeginDisabledGroup(!this._isAuthenticationRequired);
```
j9liu, line 152: I haven't been able to test the UI, so I don't know. When this button is disabled, is it visually affected, e.g. greyed out? If so, great! If not, I think that it will be a little confusing if the user tries to click it but finds it can't open.

j9liu (commented):
@azrogers Unfortunately I wasn't able to test this thoroughly because of the crash. I took an initial pass through the code and will follow-up with more thorough testing once the bug is resolved.

native~/Editor/src/CesiumIonSessionImpl.cpp
j9liu: Also it looks like this is initialized in the function body here, but is `static const` in Cesium for Unreal. Is there something preventing the two implementations from being the same?

Editor/CesiumEditorWindow.cs
azrogers: Yep, it's greyed out!

native~/Editor/src/CesiumIonSessionImpl.cpp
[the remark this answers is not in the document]
azrogers: If `_appData` doesn't have a value here, that means we never successfully connected in the first place, so we shouldn't try to resume the session, right?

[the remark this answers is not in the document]
azrogers: Reading your comment to the same effect in the Unreal PR I get what you mean. I'll change it.

j9liu:
@azrogers I'll merge this after CI passes. Thanks!

PR #464, opened 2024-06-04
Statically link MSVC runtime library

### LINKED ISSUE #419: Statically link C runtime library on Windows
We build the native code in to-be-released versions of Cesium for Unity using Visual Studio 2019 and don't explicitly specify any [run-time library options](https://learn.microsoft.com/en-us/cpp/build/reference/md-mt-ld-use-run-time-library?view=msvc-170). As a result, I believe we effectively end up using `/MD`, which means that the appropriate version of the C runtime library must be available at runtime. In other words, end users must have a "Visual C++ Redistributable" installed with a version equal to or higher than the compiler we used to build it (including patches).

Unity itself doesn't require a Visual C++ Redistributable ([source](https://support.unity.com/hc/en-us/articles/211875583-Are-there-any-required-redistributables-for-a-Unity-game)), so it's rather inconvenient that Cesium for Unity does. I think if we compile with `/MT` instead, we won't need any extra DLLs at runtime. Our built DLLs might get noticeably bigger, but it's probably worth it.

jqntn:
Hi,

Fixes #419

The built DLLs don't get much larger, maybe around 1-3MB depending on configuration.

⚠️**IMPORTANT:** This PR depends on https://github.com/CesiumGS/cesium-native/pull/900 being merged first.


kring:
Thanks for the PR @jqntn! I've pushed your changes to a branch of the same name in the main repo so that CI runs on it. Your changes look good. I'm just going to let CI build it, give the built package a quick test, and then I'll merge this.

csciguy8:
With the latest cesium-native, the official way to enable static linking is this...

`set(CESIUM_MSVC_STATIC_RUNTIME_ENABLED ON)`

... although it might work to just set `CMAKE_MSVC_RUNTIME_LIBRARY` directly too. That mentioned option was added to just make things a little easier.

jqntn:
@csciguy8 You're right.
@kring It should work fine now.

kring:
This isn't currently working:
https://github.com/CesiumGS/cesium-unity/actions/runs/9669815502/job/26677180619

I think the problem is that doing this alone is not sufficient:

```
set(CESIUM_MSVC_STATIC_RUNTIME_ENABLED ON)
```

That will make the cesium-native code use the static runtime library, but it won't affect the native code in cesium-unity itself.

csciguy8:
> This isn't currently working:

Indeed. Anything using MD_XXX (Multithreaded DLL) doesn't seem like it's getting configured like we expect.

```
Cesium3DTilesSelection.lib(ViewState.obj) : error LNK2038: mismatch detected for 'RuntimeLibrary': value 'MT_StaticRelease' doesn't match value 'MD_DynamicRelease' in Cesium.obj [D:\cesium\CesiumForUnityBuildProject\Packages\com.cesium.unity\native~\build\Runtime\CesiumForUnityNative-Runtime.vcxproj]
Cesium3DTilesSelection.lib(BoundingVolume.obj) : error LNK2038: mismatch detected for 'RuntimeLibrary': value 'MT_StaticRelease' doesn't match value 'MD_DynamicRelease' in Cesium.obj [D:\cesium\CesiumForUnityBuildProject\Packages\com.cesium.unity\native~\build\Runtime\CesiumForUnityNative-Runtime.vcxproj]
Cesium3DTilesSelection.lib(TileContent.obj) : error LNK2038: mismatch detected for 'RuntimeLibrary': value 'MT_StaticRelease' doesn't match value 'MD_DynamicRelease' in Cesium.obj [D:\cesium\CesiumForUnityBuildProject\Packages\com.cesium.unity\native~\build\Runtime\CesiumForUnityNative-Runtime.vcxproj]
```

kring:
Ok, I pushed a change to set `CMAKE_MSVC_RUNTIME_LIBRARY` as well.

csciguy8:
To be help out this PR a bit, I tested the CI artifact produced from the latest build on this branch.

In the Unity Editor, everything seemed to work fine. I added it via the "Package Manager" window using "Add package from disk".

Unfortunately, I saw errors when running "File -> Build and Run" though. But to be fair, I saw the exact same errors in main as well. I don't run this very often so not sure if I'm doing wrong or it's a new problem that was introduced for this release.

`Shader error in 'Cesium/CesiumDefaultTilesetShader': maximum ps_5_0 sampler register index (16) exceeded at line 428 (on d3d11)`
`Shader error in 'Cesium/CesiumDefaultTilesetShader': maximum ps_4_0 sampler register index (16) exceeded at line 6450 (on d3d11)`

kring:
Build and Run is working fine in this branch on my system @csciguy8. Were you building a clean checkout of Cesium for Unity Samples? I'm going to merge this because I think it's extremely important for this week's release, but we should dig into what's going wrong on your system, because others might have similar problems.

kring:
Thanks again @jqntn!

PR #465, opened 2024-06-06
Support custom ellipsoids in Unity

azrogers:
In line with CesiumGS/cesium-unreal#1432 and CesiumGS/cesium-native#890, this change lets the Unity runtime support custom ellipsoids by creating ScriptableObjects for them and specifying those assets on CesiumGeoreference components.


csciguy8:
This looks like a nice added feature. Can we get mention of it in CHANGES.md?

Also, it's not obvious to me what the typical workflow is to make a new custom ellipsoid, maybe that can be mentioned as well?

(I left this same comment on the cesium-unreal version of this PR too)

WGS84.asset
csciguy8, line 1: Verified this asset looks good in cesium-unity-samples, running with Unity 2021.3.10f1

csciguy8:
Playing with the ellipsoids, found this odd side effect.

All I did was click on the Ellipsoid Override. Selected WGS84, then selected None. The terrain refreshed and had this weird squashed appearance.

image

Runtime/CesiumCartographicPolygon.cs
```diff
@@ -138,7 +138,7 @@ internal List<double2> GetCartographicPoints(Matrix4x4 worldToTileset)
                 float3 worldPosition = knot.Transform(localToWorld).Position;
                 float3 unityPosition = worldToTileset.MultiplyPoint3x4(worldPosition);
                 double3 ecefPosition = georeference.TransformUnityPositionToEarthCenteredEarthFixed(unityPosition);
-                double3 cartographicPosition = CesiumWgs84Ellipsoid.EarthCenteredEarthFixedToLongitudeLatitudeHeight(ecefPosition);
+                double3 cartographicPosition = georeference.ellipsoid.EllipsoidCenteredEllipsoidFixedToLongitudeLatitudeHeight(ecefPosition);
```
csciguy8, line 141: `EllipsoidCenteredEllipsoidFixedToLongitudeLatitudeHeight` is really wordy.

Trying to think of a way to help this, would `CenteredFixedToLongitudeLatitudeHeight` by any better?

csciguy8:
Found another odd behavior...

If I open up a sample scene, then switch the ellipsoid to WGS84, after the tileset refreshes, some tiles are missing. These tiles loaded just fine before I switched the ellipsoid...

image

native~/extern/cesium-native
csciguy8, line 1: It looks like we can update this all the way to June 20th (the last of the related ellipsoid commits).

This cesium-native update is only going to Jun 14th.

Runtime/CesiumSimplePlanarEllipsoidCurve.cs
```diff
@@ -22,10 +22,10 @@ public partial class CesiumSimplePlanarEllipsoidCurve
         /// A <see cref="CesiumSimplePlanarEllipsoidCurve"/> if a curve can successfully be 
         /// created between the two points, or null otherwise.
         /// </returns>
-        public static CesiumSimplePlanarEllipsoidCurve FromEarthCenteredEarthFixedCoordinates(double3 sourceEcef, double3 destinationEcef)
+        public static CesiumSimplePlanarEllipsoidCurve FromEarthCenteredEarthFixedCoordinates(CesiumEllipsoid ellipsoid, double3 sourceEcef, double3 destinationEcef)
```
csciguy8, line 25: I'm guessing the mention of "Earth" in this function name is no longer valid?

Runtime/CesiumSimplePlanarEllipsoidCurve.cs
```diff
@@ -70,8 +70,8 @@ public static CesiumSimplePlanarEllipsoidCurve FromLongituteLatitudeHeight(doubl
         /// <returns>The position of the given point on this curve in Earth-Centered, Earth-Fixed coordinates.</returns>
         public partial double3 GetPosition(double percentage, double additionalHeight = 0.0);
 
-        private partial bool CreateFromEarthCenteredEarthFixedCoordinates(double3 sourceEcef, double3 destinationEcef);
-        private partial bool CreateFromLongitudeLatitudeHeight(double3 sourceLlh, double3 destinationLlh);
+        private partial bool CreateFromEarthCenteredEarthFixedCoordinates(CesiumEllipsoid ellipsoid, double3 sourceEcef, double3 destinationEcef);
```
csciguy8, line 73: I'm guessing the mention of "Earth" in this function name is no longer valid?

native~/Runtime/src/CameraManager.cpp
```diff
@@ -71,13 +73,19 @@ ViewState unityCameraToViewState(
   double horizontalFOV =
       2 * glm::atan(camera.aspect() * glm::tan(verticalFOV * 0.5));
 
+  const CesiumGeospatial::Ellipsoid& ellipsoid =
+      georeference != nullptr
```
csciguy8, line 77: The nullptr check isn't needed here. `georeference` is a const reference, not a pointer.

native~/Runtime/src/UnityTilesetExternals.cpp
csciguy8, line 1: I think you once had changes in this file, but ended up not needing them. Can revert all these.

We never actually store the ellipsoid in `TilesetExternals`

native~/Runtime/src/CesiumGlobeAnchorImpl.cpp
```diff
@@ -174,7 +192,8 @@ void CesiumGlobeAnchorImpl::SetLocalToEastUpNorthRotation(
   GlobeAnchor cppAnchor(
       UnityTransforms::fromUnity(anchor._localToGlobeFixedMatrix()));
 
-  LocalHorizontalCoordinateSystem eastUpNorth = createEastUpNorth(cppAnchor);
+  LocalHorizontalCoordinateSystem eastUpNorth =
+      createEastUpNorth(cppAnchor, getAnchorEllipsoid(anchor));
```
csciguy8, line 196: Can we call `getAnchorEllipsoid` just once here?

I'm assuming we are expecting the same ellipsoid between the first and second call.

native~/Runtime/src/CesiumBingMapsRasterOverlayImpl.cpp
```diff
@@ -81,7 +81,7 @@ void CesiumBingMapsRasterOverlayImpl::AddToTileset(
       overlay.bingMapsKey().ToStlString(),
       mapStyle,
       "",
-      CesiumGeospatial::Ellipsoid::WGS84,
+      tileset.NativeImplementation().getTileset()->getEllipsoid(),
```
csciguy8, line 84: Could replace this with `pTileset->getEllipsoid`, since `pTileset` has already been checked against null.

csciguy8 (changes requested):
Thanks @azrogers! Looks like a lot of good changes to support the new functionality.

Requesting changes for some of the odd behavior that I tested.

Also a handful of code comments to consider too.

native~/Runtime/src/CameraManager.cpp
azrogers: Unity objects, including CesiumGeoreference, actually overload `!= nullptr` to work as an existence check. You can see it elsewhere in the file!

csciguy8: Ah!... not sure how I feel about that... but thanks for the explanation.

azrogers:
There's a failing test with CesiumGlobeAnchor I'm still working on, but other than that I think I've updated everything.

azrogers:
@csciguy8 should be ready for another look now!

csciguy8 (approved):
Looks great @azrogers !

Will merge once the final CI run completes.

Did encounter a minor refresh issue during testing. If I used a custom ellipsoid, then edited it, the tileset wouldn't refresh. I would have to manually click "Refresh Tileset". Not really sure if this is fixable, and doesn't seem substantial enough to block this PR.

native~/Runtime/src/CameraManager.cpp
kring: The way to think about it is that instances of the generated classes in the DotNet namespace act like C# references. And just like it makes sense to compare a C# reference to null, it makes sense to compare an instance of these wrapper classes to nullptr. It makes no sense to have a stack-allocated instance of a C# object because C# classes are - by definition - heap allocated. So an instance of these classes really can't be anything other than a reference.

csciguy8: It makes practical sense for sure. I think I was just caught a little off-guard since I was reviewing a .cpp file, and had my C++ glasses on.

PR #467, opened 2024-06-12
Add support for ellipsoid tileset loader

jqntn:
Supports https://github.com/CesiumGS/cesium-native/pull/908 by providing a new "From WGS84 Ellipsoid" option in the `Cesium3DTileset`'s "Tileset Source" dropdown.

image


azrogers:
Thank you @jqntn!

PR #470, opened 2024-06-14
Fix token selector panel problems

### LINKED ISSUE #469: Cesium ion token set on "Token" panel is not saved
This has been reported several times on the community forum recently:
* https://community.cesium.com/t/imagery-not-loading-in-cesium-for-unity-in-android/32979/2?u=kevin_ring
* https://community.cesium.com/t/help-with-project-default-tokens/33071

And I just confirmed it myself.

If you edit the token field in the `CesiumSettings/Resources/CesiumIonServers/ion.cesium.com` asset directly using the project tab, Unity silently saves it when you either exit the editor or save the project.

However, if you edit the token using Cesium's "Token" panel, then the token is never saved.

This is major problem that needs to be fixed urgently, but as a workaround you can get the token to save by clicking the folder icon next to the ion server selector to navigate to the asset in the Project tab:
image

Then, with the server asset selected, in the Inspector tab, change any field and then change it back. Now when you exit the Unity Editor or save the project, the new token should be saved.

kring:
Fixes #469 

Also fixed up some code that was assigning by-value returns from a function to references (setting up a likely use-after-free, though I didn't see any actual problems as a result of this). And some other code that was attempting to std::move away `CesiumIonClient::Token` records owned by someone else (luckily the source was const so the std::move was only misleading, not harmful).

The state tracking on the Tokens panel was also a little bit broken, and I hopefully fixed it up. As a reminder, the panel looks like this:

image

This panel is a little confusing because it has two purposes and they are sometimes slightly conflicting. First, it should show the user how the project default token is _currently_ configured for their project. Second - particularly when no token is selected yet - it should make it easy for the user to choose a suitable project default token.

Take a look at the changed code and the comments to understand how I think this should all come together. Previously the panel could pop up with "Use an existing token" checked and some token selected in the dropdown below it, while _actually_ using some totally different token displayed in the Specify a token section.

### DIFF
--- a/native~/Editor/src/SelectIonTokenWindowImpl.cpp
+++ b/native~/Editor/src/SelectIonTokenWindowImpl.cpp
@@ -15,6 +15,7 @@
 #include <DotNet/System/Array1.h>
 #include <DotNet/System/Collections/Generic/List1.h>
 #include <DotNet/System/Object.h>
+#include <DotNet/UnityEditor/EditorUtility.h>
 #include <DotNet/UnityEngine/Debug.h>
 #include <DotNet/UnityEngine/GameObject.h>
 #include <DotNet/UnityEngine/Object.h>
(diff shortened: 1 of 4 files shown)


native~/Editor/src/SelectIonTokenWindowImpl.cpp
```diff
@@ -274,6 +303,7 @@ void updateDefaultToken(
 
     server.defaultIonAccessToken(System::String(response.value->token));
     server.defaultIonAccessTokenId(System::String(response.value->id));
+    UnityEditor::EditorUtility::SetDirty(server);
 
     // TODO: source control
```
j9liu, line 308: I think I left this comment in a long while ago and forgot to take it out :sweat_smile: will do that now

j9liu (approved):
Looks great @kring ! I verified that the token saves + the UI more accurately reflects the correct token state. Much of this code was derived from Cesium for Unreal, so it might be good to (eventually) port these changes over to address the same issues.

I'm going to push some minor changes and merge.

PR #479, opened 2024-07-25
Use vcpkg version of cesium-native

### LINKED ISSUE #432: Crash in `CesiumIonSessionImpl` during Unity code reloading
Same issue I wrote about on the forums: https://community.cesium.com/t/crash-in-cesium-for-unity-during-code-reloading/29544

Occasionally during code reloading, Cesium for Unity will crash due to a read access violation coming from one of two lines, both in CesiumIonSessionImpl.cpp. First, on line 266:
```
// Verify that the connection actually works.
pConnection->me()
    .thenInMainThread(
        [this, pConnection](
            CesiumIonClient::Response<CesiumIonClient::Profile>&& response) {
        logResponseErrors(response);
        if (response.value.has_value()) {
            this->_connection = std::move(*pConnection); // CRASH HERE
        }
        this->_isResuming = false;
        this->_quickAddItems = nullptr;
        this->broadcastConnectionUpdate();

        this->startQueuedLoads();
    })
    .catchInMainThread([this](std::exception&& e) {
        logResponseErrors(e);
        this->_isResuming = false;
    });
```
And second on line 383:
```
this->_connection->me()
    .thenInMainThread(
        [this](
            CesiumIonClient::Response<CesiumIonClient::Profile>&& profile) {
        this->_isLoadingProfile = false;
        this->_profile = std::move(profile.value);
        this->broadcastProfileUpdate();
        if (this->_loadProfileQueued)
            this->refreshProfile();
    })
    .catchInMainThread([this](std::exception&& e) {
        this->_isLoadingProfile = false;
        this->_profile = std::nullopt; // CRASH HERE
        this->broadcastProfileUpdate();
        if (this->_loadProfileQueued)
            this->refreshProfile();
    });
```

These crashes seem to happen when one of the above lines runs after the `CesiumIonSessionImpl` object has been destroyed. I believe this happens when two code reloads happen in quick succession after one another - a new session is created after one code reload only to be destroyed by the start of the second, before the callback has finished. Unfortunately, putting `session` into the captures of the two lambdas above doesn't seem to fix the issue.

I've been able to get this crash to happen using this edit-mode test:
```
using System.Collections;
using System.Collections.Generic;
using System.Linq;
using UnityEditor;
using UnityEditor.Compilation;
using UnityEngine;
using UnityEngine.TestTools;

public class TestCesiumIonSession
{
    [UnityTest]
    public IEnumerator CrashWhenResumingSession()
    {
        EditorUserBuildSettings.SwitchActiveBuildTarget(BuildTargetGroup.Standalone, BuildTarget.StandaloneWindows);

        yield return new EnterPlayMode(true);
        yield return new ExitPlayMode();

        CompilationPipeline.RequestScriptCompilation();
        EditorUserBuildSettings.SwitchActiveBuildTarget(BuildTargetGroup.Android, BuildTarget.Android);
        yield return null;

        yield return new EnterPlayMode(true);
        yield return new ExitPlayMode();

        EditorUserBuildSettings.SwitchActiveBuildTarget(BuildTargetGroup.Standalone, BuildTarget.StandaloneWindows);
    }
}
```

I'm not sure how many of these lines are needed, but this seems to make the crash happen about 2/3rds of the time when run. Sometimes it takes a few repeated runs to happen. It seems to happen more often when Unity is compiling in the background than when focused in the foreground, but I might just be imagining this.

kring:
There's a lot going on this PR. Here's a summary.

### vcpkg

We're now using the vcpkg-ified version of cesium-native from CesiumGS/cesium-native#820.

### Unity 2022.3.41f.

Previously we used 2021.3.13f1. This means we can actually compile and test all of our code, including `CesiumCartographicPolygon` which requires 2022.3. The plugin will still work with 2021.3 after building with 2022.3, but there is a chance we might miss compatibility problems with the older release. 2021.3 will be out of support as soon as Unity 6 is released, though, so it should be a very short-term problem.

This is probably a good change in general, but the immediate motivation for making the switch here was that 2021.3 uses an older version of the Android NDK that uses Clang 9, which doesn't have very good C++20 support. 2022.3 uses Clang 12 on Android, which is much better.

To use the new Unity version to do builds, I had to fix a longstanding bug where the build process was baking-in _all_ code that came out of a Roslyn code generator. This was fine in 2021.3 because Reinterop was the only Roslyn code generator. In 2022.3 this is no longer true, and it was breaking the build.

### C++ 20

The C++ code in Cesium for Unity (but not cesium-native) now target C++20. This is necessary to use [swl-variant](https://github.com/groundswellaudio/swl-variant).

### swl-variant

This is an alternate `variant` implementation that requires C++20 and has _much_ better compile-time performance in terms of both CPU time and memory usage. Like I did with [Unreal](https://github.com/CesiumGS/cesium-unreal/pull/1447), I've only switched the variants in cesium-unity itself to use this implementation. cesium-native still uses `std::variant`, though it may be worthwhile to change that too at some point.

### macOS builds now run on macOS 14 / ARM64

Previously we were stuck on macOS 12 and x64. With the reduced compile-time memory usage from switching to swl-variant, we're no longer prevented from using GitHub's Apple Silicon runners (which are weirdly memory constrained). This gets us onto the more modern platform (OS and processor architecture) and speeds up our builds.

### Builds are faster

In main, builds took over 2 hours on Windows and almost an hour and a half on macOS. In this branch, total CI times on both platforms are cut approximately in half. This is a result of swl-variant, faster runners, and caching builds of third-party dependencies via vcpkg. I also modified the build process to immediately exit Unity after the native code finished building, saving a bunch of pointless work building a Player application that we don't actually need.

The improved build times are also a result of disabling Unity's audio system during our build. Incredibly, on macOS, Unity was doing...something...with audio for 20 minutes every time it started up, presumably because the GitHub runners can't actually do audio or somesuch. Disabling it massively reduced our macOS build times.

### Native code build logs are uploaded as build artifacts

Previously, if the native code failed to compile on one of our many platforms, it was hard to figure out what went wrong. We had to add a bit of script to dump the log to the build output, and run the build again. Now, a build artifact is created with all of the native build logs, so they can easily be inspected if needed.

### Consistent macOS and iOS version targetting

Previously, we were targeting macOS 12.7 in our Editor build of the plugin. And 10.13 for Player builds. Now, we're targetting 10.15 for both. We moved from 10.13 to 10.15 because 10.13 doesn't support `std::filesystem::path`.

~On iOS, we previously weren't attempting to specify a target version, so we were getting xcode's default (whatever that is). Now we're explicitly targetting iOS 12, which is the minimum version supported by Unity 2022.3.~ Update: While we weren't explicitly specifying a target before, we were (perhaps coincidentally) getting iOS 12 anyway.

### Floating point comparisons

When we started running the tests on an Apple Silicon Mac, we initially had some test failures. Apple's ARM processors seem to be a bit less tolerant of testing floating point equality compare to x64 processors. So I added some epsilons to some comparisons in tests. I also changed `CesiumFlyToController.DetectMovementInput` to only allow movement of more than 1/1000 of a meter to cancel an in-progress flight.

### CesiumIonSession crash

We have sometimes seen crashes related to CesiumIonSession. It became quite consistent when running the tests in Unity 2022.3. The crash was caused by a race condition where the managed `CesiumIonSession` was garbage collected (perhaps due to AppDomain unload) while it was still attempting to resume a previous session, resulting in a use-after-free. I fixed this by holding a reference to the managed object while requests are in flight, and checking when they complete that the managed object is still valid.

Fixes #432 
(I think!)


kring:
I'm struggling with the Android build, specifically with installing OpenSSL from vcpkg. Unity includes the Android NDK, which is installed to `C:\Program Files\Unity\Hub\Editor\2021.3.13f1\Editor\Data\PlaybackEngines\AndroidPlayer\NDK` by default. That path has a space in it, and OpenSSL's build process apparently can't handle that. It says:

```
-- Warning: Tools with embedded space may be handled incorrectly by configure:
   C:/Program Files/Unity/Hub/Editor/2021.3.13f1/Editor/Data/PlaybackEngines/AndroidPlayer/NDK/toolchains/llvm/prebuilt/windows-x86_64/bin/clang.exe
   C:/Program Files/Unity/Hub/Editor/2021.3.13f1/Editor/Data/PlaybackEngines/AndroidPlayer/NDK/toolchains/llvm/prebuilt/windows-x86_64/bin/clang++.exe
   C:/Program Files/Unity/Hub/Editor/2021.3.13f1/Editor/Data/PlaybackEngines/AndroidPlayer/NDK/toolchains/llvm/prebuilt/windows-x86_64/bin/llvm-nm.exe
   C:/Program Files/Unity/Hub/Editor/2021.3.13f1/Editor/Data/PlaybackEngines/AndroidPlayer/NDK/toolchains/llvm/prebuilt/windows-x86_64/bin/aarch64-linux-android-strip.exe
   C:/Program Files/Unity/Hub/Editor/2021.3.13f1/Editor/Data/PlaybackEngines/AndroidPlayer/NDK/toolchains/llvm/prebuilt/windows-x86_64/bin/llvm-objdump.exe
   C:/Program Files/Unity/Hub/Editor/2021.3.13f1/Editor/Data/PlaybackEngines/AndroidPlayer/NDK/toolchains/llvm/prebuilt/windows-x86_64/bin/aarch64-linux-android-ranlib.exe
   C:/Program Files/Unity/Hub/Editor/2021.3.13f1/Editor/Data/PlaybackEngines/AndroidPlayer/NDK/toolchains/llvm/prebuilt/windows-x86_64/bin/aarch64-linux-android-ar.exe
   C:/Program Files/Unity/Hub/Editor/2021.3.13f1/Editor/Data/PlaybackEngines/AndroidPlayer/NDK/toolchains/llvm/prebuilt/windows-x86_64/bin/ld.lld.exe
-- Warning: Arguments with embedded space may be handled incorrectly by configure:
   --gcc-toolchain=C:/Program Files/Unity/Hub/Editor/2021.3.13f1/Editor/Data/PlaybackEngines/AndroidPlayer/NDK/toolchains/llvm/prebuilt/windows-x86_64
   --sysroot=C:/Program Files/Unity/Hub/Editor/2021.3.13f1/Editor/Data/PlaybackEngines/AndroidPlayer/NDK/toolchains/llvm/prebuilt/windows-x86_64/sysroot
```

And then fails later with:

```
/bin/sh: line 1: C:/Program: No such file or directory
make[1]: *** [Makefile:4646: crypto/aes/libcrypto-lib-aes_cbc.o] Error 127
```

Usually a good strategy when dealing with broken software that can't handle spaces in paths is to use the old school 8.3 version of the path. So I tried setting the `ANDROID_NDK_ROOT` environment variable to `C:/Progra~1/Unity/Hub/Editor/2021.3.13f1/Editor/Data/PlaybackEngines/AndroidPlayer/NDK`. No more space. But that still doesn't work, because CMake "helpfully" rewrites it to the long path so it breaks in the same way. 🙄 CMake bug report here:
https://gitlab.kitware.com/cmake/cmake/-/issues/25639

The original author closed it without a fix in CMake after they worked around it with some complicated code to create NTFS junction points. I'll probably end up doing the same.

By the way, this wasn't an issue before vcpkg because we avoided the need for an actual OpenSSL dependency by carving out the bits of s2geometry we actually need and leaving out the bits that depend on OpenSSL. Doing the same with vcpkg would be...tricky.

kring:
Today's problem is that the Android build is failing only on CI. The reason appears to be that the GitHub runner has a recent version (v27) of the Android NDK installed, and the `ANDROID_NDK_ROOT` environment variable set accordingly. Because this environment variable is set, Unity doesn't override it with its own (older) copy of the Android NDK.

This alone wouldn't be a problem, except for this issue:
https://github.com/android/ndk/issues/2032#issuecomment-2267532582

The linked issue says this will be addressed in NDK v28, but it's not completely clear this will fix the problem for us. Recent versions of the Android NDK's CMake toolchain require CMake 3.6, where's asyncplusplus only requires CMake 3.1. We're of course using a plenty-new CMake, but CMake has a whole policy system whereby newer versions behave like older ones based on the declared required version of the project. In any case, I'm going to try to force our builds to use Unity's copy of the NDK, which should fix the problem.

native~/Editor/src/CesiumIonSessionImpl.cpp
```diff
@@ -758,11 +785,18 @@ CesiumAsync::Future<bool> CesiumIonSessionImpl::ensureAppDataLoaded(
              this->_pAssetAccessor,
              server.apiUrl().ToStlString())
       .thenInMainThread(
-          [this, session](
+          [this, session, asyncSystem = this->_asyncSystem](
               CesiumIonClient::Response<CesiumIonClient::ApplicationData>&&
                   appData) {
             CesiumAsync::Promise<bool> promise =
-                this->_asyncSystem.createPromise<bool>();
+                _asyncSystem.createPromise<bool>();
```
j9liu, line 792: Hm, this lambda captures `asyncSystem` but uses `_asyncSystem` in the promise. Seems unintentional?

.gitmodules
```diff
@@ -6,4 +6,8 @@
 	url = https://github.com/htacg/tidy-html5.git
 [submodule "native~/extern/enum-flags"]
 	path = native~/extern/enum-flags
-	url = https://github.com/grisumbras/enum-flags.git
+	url = https://github.com/kring/enum-flags.git
+[submodule "native~/extern/swl-variant"]
+	path = native~/extern/swl-variant
+	url = https://github.com/kring/swl-variant.git
```
j9liu, line 12: Just wanted to check, are we using these forks to avoid future updates to the original repos?

Editor/CompileCesiumForUnityNative.cs
```diff
@@ -527,17 +550,29 @@ private static void ConfigureEnvironmentVariables(IDictionary<string, string> en
...
-                if (!string.IsNullOrEmpty(ndkRoot))
-                {
-                    string make = Path.Combine(ndkRoot, "prebuilt", "windows-x86_64", "bin", "make.exe").Replace('\\', '/');
-                    library.ExtraConfigureArgs.Add($"-DCMAKE_MAKE_PROGRAM=\"{make}\"");
-                }
+                // if (!string.IsNullOrEmpty(ndkRoot))
+                // {
+                //     string make = Path.Combine(ndkRoot, "prebuilt", "windows-x86_64", "bin", "make.exe").Replace('\\', '/');
+                //     library.ExtraConfigureArgs.Add($"-DCMAKE_MAKE_PROGRAM=\"{make}\"");
+                // }
```
j9liu, line 575: Delete?

native~/Runtime/src/CesiumMetadataValueImpl.cpp
```diff
@@ -55,21 +56,22 @@ namespace CesiumForUnityNative {
...
+template <typename F>
+auto getNativeBooleanValue(
+    const DotNet::CesiumForUnity::CesiumMetadataValue& value,
+    F&& callback) {
   std::optional<bool> maybeBoolean =
       CesiumForUnity::CesiumMetadataValue::GetObjectAsBoolean(
           value.objectValue());
   if (maybeBoolean) {
-    return *maybeBoolean;
+    return callback(*maybeBoolean);
```
j9liu, line 67: The other `get` functions use `std::move()` but this one omits it, so I just wanted to check if that was intended. (Perhaps a boolean is trivial enough that it's not necessary?)

native~/Runtime/src/CesiumMetadataValueImpl.cpp
```diff
@@ -305,824 +393,771 @@ getNativeMatValue(const DotNet::System::Object& object) {
...
   DotNet::System::String string =
       CesiumForUnity::CesiumMetadataValue::GetObjectAsString(
           value.objectValue());
   if (string == nullptr) {
-    return std::monostate();
+    return callback(swl::monostate());
   }
 
-  return string.ToStlString();
+  return callback(string.ToStlString());
```
j9liu, line 566: Same here, is `std::move` omitted on purpose?

j9liu (changes requested):
@kring looks great! I won't pretend like I completely understand CMake stuff, but everything else looks good to me.

I left some comments, and could you also update CHANGES.md + resolve the conflicts with main? (I'm worried if I do the latter I'll mess something up 😅 )

I'd also like to merge the cesium-native PR before we merge this one, but I didn't see that one marked as ready.

native~/Editor/src/CesiumIonSessionImpl.cpp
kring: Yes! Good catch!

.gitmodules
kring: No, we are using forks because we had to fix bugs in those two libraries. I should move them to CesiumGS at least, though I was hoping in both cases that the upstream would accept my PRs quickly and we wouldn't need to be on a fork for long.

native~/Runtime/src/CesiumMetadataValueImpl.cpp
kring: Right, doesn't make any sense to move something that's known to be a bool.

native~/Runtime/src/CesiumMetadataValueImpl.cpp
kring: That's already an rvalue (because ToStlString returns by value), no need for move there.

.gitmodules
kring: Here are the PRs, for reference (I only opened the first of these myself):
* https://github.com/groundswellaudio/swl-variant/pull/5
* https://github.com/grisumbras/enum-flags/pull/32

kring:
Thanks for reviewing @j9liu, this is ready for another look.

kring:
Final testing across all platforms:

❌ means I cannot test because I don't have access to the necessary hardware.

* Editor
  * [ ] Windows x64
  * [ ] macOS x64
  * [ ] ❌ macOS ARM64
* Build Game from Windows
  * [ ] Windows x64
  * [ ] Android x64
  * [ ] Android ARM64
  * [ ] Universal Windows Platform x64
  * [ ] Universal Windows Platform ARM64
* Build Game from macOS
  * [ ] macOS Universal
  * [ ] iOS ARM64
* Run Game
  * [ ] Windows x64
  * [ ] macOS Universal on x64
  * [ ] ❌ macOS Universal on ARM64
  * [ ] ❌ iOS ARM64
  * [ ] ❌ Android x64
  * [ ] Android ARM64 (tested on Pixel 6 Pro)
  * [ ] Universal Windows Platform x64
  * [ ] ❌ Universal Windows Platform ARM64

kring:
I've done all the platform testing I can here, too, so this is ready.

j9liu:
Thanks @kring ! Merging now 😎

Issue #476, opened 2024-07-08
Game-breaking bug in v1.11.0


electrum-bowie:
I'm not at my computer right now, but I'm just making sure that you guys are aware, I don't have the error logs because I instantly downgraded to the previous one and I didn't have time to do anything else that day either way.

The error logs had to do with something related to 'ellipsoid', and each time I tried to write to the 'longitudeLatitudeHeight' property the code failed with that error in the console, completely breaking my project...

electrum-bowie:
If you aren't aware of this, I'll continue providing info here when I get the chance

j9liu:
Hey @electrum-bowie, thanks for letting us know. Can you provide step-by-step instructions for reproducing the error? It would help us to have a minimal reproducible example.

electrum-bowie:
> The error logs had to do with something related to 'ellipsoid', **and each time I tried to write to the 'longitudeLatitudeHeight' property the code failed with that error in the console**, completely breaking my project...

azrogers:
@electrum-bowie Do you mean setting the `longitudeLatitudeHeight` property on a `CesiumGlobeAnchor`? I've tried setting this property in the latest version of Cesium for Unity and I'm not able to reproduce this bug. Could you share a sample of the code you were using that was causing this issue?

electrum-bowie:
Through code, of course setting it through the UI would work

azrogers:
Setting it through the inspector calls the same method as setting it from code, so I'm not sure it's "of course." 😉 Regardless, I tested with a script. This is the script I used:
```
using CesiumForUnity;
using System.Collections;
using System.Collections.Generic;
using Unity.Mathematics;
using UnityEngine;

public class Test : MonoBehaviour
{
    // Update is called once per frame
    void Update()
    {
        var anchor = GetComponent<CesiumGlobeAnchor>();
        anchor.longitudeLatitudeHeight = double3.zero;
    }
}
```

If you run this in your project on Cesium 1.11.0, does this cause an issue?

azrogers:
@electrum-bowie Are you still experiencing this issue?

electrum-bowie:
seems like the issue is now gone

electrum-bowie:
Nope, here it is:
![image](https://github.com/user-attachments/assets/cffb1e07-9e95-4156-9b21-7281b825c236)
![Uploading image.png…]()

electrum-bowie:
after running:

```cs
[...]

PR #480, opened 2024-07-25
Fix changing CesiumGlobeAnchor position while disabled

azrogers:
As noted in #476, the ellipsoid changes added a reference to `_georeference` in the `longitudeLatitudeHeight` property on `CesiumGlobeAnchor`. This is fine for most purposes, as `_georeference` will be set by the time `longitudeLatitudeHeight` is used - unless the `CesiumGlobeAnchor` is disabled. This change just makes sure the georeference has been found before trying to access it.

### DIFF
--- a/Runtime/CesiumGlobeAnchor.cs
+++ b/Runtime/CesiumGlobeAnchor.cs
@@ -188,11 +188,25 @@ namespace CesiumForUnity
             get
             {
                 this.UpdateGeoreferenceIfNecessary();
+
+                if (this._georeference == null || this._georeference.ellipsoid == null)
+                {
+                    // Cannot get property if there is no georeference, or the georeference has no ellipsoid.
+                    return new double3(0.0, 0.0, 0.0);
+                }
+
                 return this._georeference.ellipsoid.CenteredFixedToLongitudeLatitudeHeight(this.positionGlobeFixed);
             }
             set
             {
                 this.UpdateGeoreferenceIfNecessary();
+
+                if (this._georeference == null || this._georeference.ellipsoid == null)
+                {
+                    // Cannot set property if there is no georeference, or the georeference has no ellipsoid.
+                    return;
+                }
+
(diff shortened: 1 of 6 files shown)


Runtime/CesiumGlobeAnchor.cs
```diff
@@ -185,9 +185,14 @@ public double4x4 localToGlobeFixedMatrix
         /// </remarks>
         public double3 longitudeLatitudeHeight
         {
-            get => this._georeference.ellipsoid.CenteredFixedToLongitudeLatitudeHeight(this.positionGlobeFixed);
+            get
+            {
+                this.UpdateGeoreferenceIfNecessary();
+                return this._georeference.ellipsoid.CenteredFixedToLongitudeLatitudeHeight(this.positionGlobeFixed);
```
kring, line 191: What if this CesiumGlobeAnchor isn't inside a CesiumGeoreference? In that case, `_georeference` will still be null after the `UpdateGeoreferenceIfNecessary` call. Obviously the globe anchor can't actually work in that scenario, but it shouldn't crash, either. Returning (0,0,0) would probably be fine.

kring:
Looks good other than that one thing. Please remember to update CHANGES.md, too.

j9liu:
@kring Ashley probably won't be able to get back to this before the release (since she's out).

Would it be okay if you or @csciguy8 want to implement the feedback + merge this to main? (I will also be out, but can try to support if needed)

kring:
Yeah good call @j9liu. I'll make that small change and merge it.

electrum-bowie:
???

electrum-bowie:
@kring when will i be able to update the cesium unity package to receive this fix ?

kring:
We're planning to release it tomorrow. You can grab a build from CI in the meantime if you like.

electrum-bowie:
CI ?

kring:
Continuous Integration. The plugin is built on GitHub Actions on every commit. If you're not familiar, better to just wait for the release.

PR #482, opened 2024-07-31
Fix incorrect ECEF->LLH transformation in CesiumCartographicPolygon.

kring:
After #465, `CesiumCartographicPolygon` was incorrectly calling `LongitudeLatitudeHeightToCenteredFixed` when attempting to transform ECEF to Longitude/Latitude/Height. That is, it was transforming in the wrong direction. This made the vertex positions in the polygon incorrect, breaking the polygon functionality completely.

Reported here:
https://community.cesium.com/t/cesiumpolygonrasteroverlay-can-not-clip-vectortilemap-in-v1-11/34093

### DIFF
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -5,6 +5,7 @@
 ##### Fixes :wrench:
 
 - Fixed a bug that caused a `NullReferenceException` when attempting to get or set the `longitudeLatitudeHeight` property on a disabled `CesiumGlobeAnchor`.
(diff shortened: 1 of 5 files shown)


kring:
@csciguy8 can you please review and merge this one today so that we can include it in tomorrow's release?

csciguy8 (approved):
Thanks @kring for the catch and the new test.

Looks good.

PR #491, opened 2024-08-21
Add support for double-sided materials

### LINKED ISSUE #370: Support double-sided materials
As reported on the [forum](https://community.cesium.com/t/draw-gltf-with-doublesided-material/27044). We don't support double-sided glTF materials, but marking a Unity material double-sided doesn't seem to be easy from script. I'm copy pasting my thoughts from the community forum:

>I think this is something we'd have to investigate further... I can see that there's a **Render Face** option with the ability to render both front + back faces in the Material Editor. 
>
>
>
> But from a glance at Unity's [Material](https://docs.unity3d.com/ScriptReference/Material.html) API, there's no direct setting to control this. It looks like in a custom shader, you could set a `Cull` flag [manually](https://docs.unity3d.com/Manual/SL-Cull.html). But we avoid handwriting shaders because the necessary code can differ between the three render pipelines. Shader Graph is ideal for us because it abstracts that away.
>
> Maybe the double-sided behavior can be controlled in Shader Graph by setting some keyword, but I don't know what that keyword is. In the worst case scenario, it might be possible to simulate double-sided behavior from the material by adding shader graph code to flip the normal if it is facing away from the camera. And then to add a keyword to only enable this if the material is double-sided.

### LINKED ISSUE #486: CesiumPolygonRasterOverlay Cliping can not work when use HDRP
I test in unity2022.3 and Unity6000,
The CesiumPolygonRasterOverlay  has the same bug.
CesiumPolygonRasterOverlay Can not Clip when use HDRP,but URP work fine.

kring:
As reported here:
https://community.cesium.com/t/3d-tiles-model-appeared-with-face-orientation-flipped-backface-culling/34155

We previously weren't supporting glTF's [doubleSided](https://registry.khronos.org/glTF/specs/2.0/glTF-2.0.html#double-sided) property. Now we are.

 This was tricky because Unity controls this with a render-pipeline-dependent material property. I also had to add some nodes to our CesiumDefaultTilesetShader to flip the direction of the normal for back faces, as required by the glTF spec.

This is a draft because it's not quite working well in HDRP yet. For reasons I don't understand, HDRP defies all attempts to dynamically turn on double-sided rendering dynamically. I have some local changes where I turn it on via the UI, and then I'm able to turn it _off_ dynamically. But I haven't committed that yet because it seems dodgy. Also, unlike the othe render pipelines, HDRP seems to automatically invert normals for back faces, which means my nodes that do that manually are now making the normals wrong again. I'm not sure yet the best way to deal with that.

Fixes #370
Fixes #486

### DIFF
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -14,6 +14,7 @@
 
 - Fixed a bug that could cause a crash in `CesiumIonSession` when the object was garbage collected or the AppDomain was unloaded while network requests were in progress.
 - Fixed a bug that could cause `CesiumFlyToController` to unexpectedly interrupt a flight.
+- glTF `doubleSided` materials are now supported. Previously, models using this property would appear inside-out.
 
 ## v1.11.1 - 2024-08-01
 
--- a/Runtime/ConfigureReinterop.cs
+++ b/Runtime/ConfigureReinterop.cs
@@ -158,6 +158,7 @@ namespace CesiumForUnity
             int crc = meshRenderer.material.ComputeCRC();
             meshRenderer.material.SetTexture(id, texture2D);
             meshRenderer.material.SetFloat(id, 1.0f);
+            meshRenderer.material.SetFloat(id, (float)CullMode.Off);
             meshRenderer.material.SetVector(id, new Vector4());
             meshRenderer.material.DisableKeyword("keywordName");
             meshRenderer.material.EnableKeyword("keywordName");
--- a/Runtime/Resources/CesiumDefaultTilesetMaterial.mat
+++ b/Runtime/Resources/CesiumDefaultTilesetMaterial.mat
@@ -33,7 +33,7 @@ Material:
   m_LightmapFlags: 2
   m_EnableInstancingVariants: 0
   m_DoubleSidedGI: 1
-  m_CustomRenderQueue: 2475
+  m_CustomRenderQueue: 2450
   stringTagMap:
     MotionVector: User
     RenderType: TransparentCutout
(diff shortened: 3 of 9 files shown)


kring:
I had hoped to get this PR ready for final review so it can be included in next week's release, but HDRP and I still are not getting along. Will probably need to push it until next release.

kring:
Ok I think this is working well now. Double-sided rendering is working in all pipelines. The way it works is that the Shader and Material now have double-sided mode enabled. At runtime, we disable it if the glTF does not require it by setting some material parameters. This seems to work a lot better than the opposite. I'm not 100% sure why, but my theory is that double-sided rendering is a "feature" (at least in some RPs). If the shaders are compiled with support for that feature, then dynamically disabling it at runtime is fine. However, if the shaders are compiled without support for the feature at all, then enabling it dynamically is not possible.

I also fixed cartographic polygons in HDRP by enabling "alpha clipping".

kring:
I'm moving this into next week's release, but time is short so we'll move it back out if there are any concerns at all.

j9liu:
Thanks @kring ! I verified that this was working using the [TextureSettingsTest](https://github.com/KhronosGroup/glTF-Sample-Models/tree/d7a3cc8e51d7c573771ae77a57f16b0662a905c6/2.0/TextureSettingsTest) sample model.

Just to sanity check, I noticed some seemingly unrelated edits to the material? Some of them seemed like autogenerated additions, but I wasn't sure about this one. Was this done by Unity too?

image

kring:
@j9liu I didn't purposely change that, so I can only imagine Unity change it itself for some reason. I reverted it back to "Auto" though, and everything still seems to be working well, so I've pushed up that change.

kring:
Merging this one for tomorrow's release, too.

PR #493, opened 2024-08-28
Add 'True Origin' placement mode

### LINKED ISSUE #426: Add "True Origin" mode to `CesiumGeoreference`
Similar to [Unreal's](https://github.com/CesiumGS/cesium-unreal/blob/main/Source/CesiumRuntime/Public/OriginPlacement.h) georeference implementation, we should have a "True Origin" option on `CesiumGeoreference`.

j9liu:
Closes #426.

This replicates the "True Origin" placement mode we have in Cesium for Unreal. This will be helpful for users with non-georeferenced tilesets.

I rearranged the `CesiumGeoreference`'s Inspector UI so the order was more intuitive. Here's what it looks like on both modes:

| Cartographic Origin | True Origin |
| ---------------------- | ------------ |
| image | image |

### DIFF
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -9,6 +9,7 @@
 ##### Additions :tada:
 
 - Editor builds for macOS now target macOS 10.15+. Previously, macOS 12.7+ was required.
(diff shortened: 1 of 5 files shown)


j9liu:
Some UI tweaks:
- Disable "Place Origin Here" for true origin. I think it's going to be more confusing if it's made available
- Made "Origin Mode" group

image

kring:
Looks great, thanks @j9liu!

PR #495, opened 2024-08-30
Fix "True Origin" inversion issue

j9liu:
I forgot that Unity is left-handed, so I accidentally left out the y-inversion necessary to get to a left-handed coordinate system. Unfortunately I didn't catch this in #493 because I was using a symmetrical model, but it became very obvious when testing with the [TextureSettingsTest ](https://github.com/KhronosGroup/glTF-Sample-Models/tree/d7a3cc8e51d7c573771ae77a57f16b0662a905c6/2.0/TextureSettingsTest) model

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -154,6 +154,30 @@ jobs:
     # Only allow a single macOS build at a time, for Unity licensing reasons
     concurrency: mac
     steps:
+      - name: Make some more disk space
+        run: |
+          df -h
+          ls /Applications
+          brew uninstall google-chrome
+          sudo rm -rf /Users/runner/Library/Android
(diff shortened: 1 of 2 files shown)


kring:
Thanks @j9liu. I checked the math a little more closely this time, and agree it's correct now. The way I think about it is that a glTF, after it's had the Y-up to Z-up transformation applied by 3D Tiles, can be thought of as having an East-North-Up axis order. By swapping the Y and Z axes (which is what the matrix now does), we end up matching the East-Up-North axis order used by the CesiumGeoreference to map the globe into the Unity world.

I'm going to separate this from the double-sided branch so we can include it even if don't end up shipping that this release.

kring:
Closing this now that #498 is merged and this branch has no additional changes.

PR #497, opened 2024-08-31
Fix macOS build by freeing up disk space on the runner

kring:
Unity version of this PR:
https://github.com/CesiumGS/cesium-unreal/pull/1510


kring:
Merging this to fix main.

Issue #409, opened 2024-02-13
Vision Pro Reinterop build issue


mechpil0t:
When building for Vision Pro (visionOS - both device and simulator SDK) you get a Reinterop not found error. 

I assume because the Reinterop has not been compiled for visionOS?

kring:
> I assume because the Reinterop has not been compiled for visionOS?

Reinterop itself is .NET code and doesn't need to be compiled for any particular platform. However, the Cesium for Unity native code has not been compiled for the Vision Pro. You may be able to get it to work by building manually from the source in this repo (_not_ the source that is included in the shipped package; it doesn't have all the bits that are needed to compile the native code). See the [developer setup instructions](https://github.com/CesiumGS/cesium-unity/blob/main/Documentation~/developer-setup.md), though you definitely won't be able to just follow them blindly. [This thread](https://community.cesium.com/t/cesium-unity-on-apple-vision-pro/29862) on the community forum discusses the possibility of using the iOS build of the native code on Vision Pro, too, though I have no idea if that will work.

GiantSox:
I'm getting closer after building the CMake project (with a few modifications) and copying the files into the Unity-generated Xcode project... but running into one last linker error. Any hints as to which library I might be missing? :)

<img width="577" alt="Screenshot 2024-03-13 at 9 35 51 AM" src="https://github.com/CesiumGS/cesium-unity/assets/6282089/6d3b7e44-5093-430b-a285-34213c188fd7">

csciguy8:
That's a cesium native function.

Is `\CesiumIonClient\src\Connection.cpp` getting compiled in addition to the other cesium-native source files?

GiantSox:
@csciguy8 As far as I can tell, yes (the build fails if I put uncompilable code in `Connection.cpp`). `CesiumIonClient` isn't supposed to be built into its own library is it?

kring:
Not much context in that error screenshot, but the interesting thing about CesiumIonClient is that the CesiumRuntime assembly doesn't use it; only CesiumEditor does. So it's possible you're copying files into one but not the other? Just a guess.

GiantSox:
@kring The error was in Xcode when linking the project, not in Unity. Though I also don't fully understand how Reinterop works. Still not sure what I'm doing wrong. Sadly adding libCesiumIonClient.a didn't fix it :(

jaiyramanan23:
> > I assume because the Reinterop has not been compiled for visionOS?
> 
[...]

PR #502, opened 2024-09-05
WIP Added Vision Pro build support

handzlikchris:
Added Vision Pro build support, this is done by:

- getting Reinterop to build libs to `Generated-VisionOS`
- building native library with IOS configuration
- changing XCode SDK to xros (after generation, workaround)
- fixing issue where Unity would not copy generated Cesium libs to Xcode project (by setting CPU to ARM64 explicitly)
- excluding some native libs from final XCode build (workaround)

Still work in progress but should be good starting place for anyone looking to use Cesium for Unity on Vision Pro / Polyspatial.

Approach taken here contains some workarounds - happy to continue on this one but would need some guidance.

1) Cesium native is built using IOS config and XCode project is later re-targetted to xros, ideally we'd have separate xros toolchain for cmake, eg `extern/xros-toolchain.cmake` when I tried to build that it'd fail when resolving triplets
```
*** The output from the command was:
Computing installation plan...
The following packages will be built and installed:
    asyncplusplus:arm64-ios-unity@1.1#2 -- /Volumes/External/src/cesium-unity-samples-vp/Packages/com.cesium.unity/native~/vcpkg/ports/asyncplusplus
  * vcpkg-cmake:arm64-osx@2024-04-23
  * vcpkg-cmake-config:arm64-osx@2024-05-23
Additional packages (*) will be modified to complete this operation.
Detecting compiler hash for triplet arm64-osx...
Compiler found: /Applications/Xcode.app/Contents/Developer/Toolchains/XcodeDefault.xctoolchain/usr/bin/c++
Unable to determine toolchain use for arm64-ios-unity with with CMAKE_SYSTEM_NAME visionOS. Did you mean to use VCPKG_CHAINLOAD_TOOLCHAIN_FILE?
note: updating vcpkg by rerunning bootstrap-vcpkg may resolve this failure.
```

Could resolve that properly to build for VisionOS from start.

2) Needed to hardcode some generated libs to be exlcuded from build
```
"libjpeg.a",
"libwebp.a"
```
As they fail to compile (duplicated symbols) - not 100% sure if that exclusion will cause issues.

Also looks like native project for this PR installs into Plugins folder quite a lot of libraries that I've not seen built in other forum threads, notably 86 libs build for `libabsl_XXX.a` eg `libabsl_bad_any_cast_impl.a`, `libabsl_string_view.a`, `libabsl_throw_delegate.a`

### DIFF
--- a/Editor/CompileCesiumForUnityNative.cs
+++ b/Editor/CompileCesiumForUnityNative.cs
@@ -173,6 +173,18 @@ namespace CesiumForUnity
             else if (library.Platform == BuildTarget.VisionOS)
             {
                 importer.SetPlatformData(library.Platform, "CPU", "ARM64");
+
+                var projectPath = Application.dataPath.Replace("Assets", "");
+                foreach (var libFilePath in Directory.EnumerateFiles(library.InstallDirectory, "*.a", SearchOption.AllDirectories))
+                {
+                    var libRelativeFilePath = libFilePath.Replace(projectPath, "");
+                    var libPluginImporter = AssetImporter.GetAtPath(libRelativeFilePath) as PluginImporter;
+                    
+                    if (libPluginImporter != null)
+                    {
+                        libPluginImporter.SetPlatformData(library.Platform, "CPU", "ARM64");
+                    }
+                }
             }
         }
 
--- a/Editor/CompileCesiumForUnityNative.cs
+++ b/Editor/CompileCesiumForUnityNative.cs
@@ -502,9 +502,16 @@ namespace CesiumForUnity
                     if (IsVisionOS(library.PlatformGroup, library.Platform))
                     {
                         var xcodeProjectPath = Path.Combine(library.BuildDirectory, "CesiumForUnityNative.xcodeproj/project.pbxproj");
-                        var originalXcodeContents = File.ReadAllText(xcodeProjectPath);
-                        var xcodeContentsTargetedToXros = originalXcodeContents.Replace("SDKROOT = iphoneos;", "SDKROOT = xros;");
-                        File.WriteAllText(xcodeProjectPath, xcodeContentsTargetedToXros);
+                        if (File.Exists(xcodeProjectPath))
+                        {
+                            var originalXcodeContents = File.ReadAllText(xcodeProjectPath);
+                            var xcodeContentsTargetedToXros = originalXcodeContents.Replace("SDKROOT = iphoneos;", "SDKROOT = xros;");
+                            File.WriteAllText(xcodeProjectPath, xcodeContentsTargetedToXros);
+                        }
+                        else
+                        {
+                            UnityEngine.Debug.Log("Xcode project does not exist, unable to change target to VisionOs");
+                        }
                     }
                     
                     args = new List<string>()
--- a/Editor/CompileCesiumForUnityNative.cs
+++ b/Editor/CompileCesiumForUnityNative.cs
@@ -377,9 +377,7 @@ namespace CesiumForUnity
                     library.ExtraConfigureArgs.Add("-DCMAKE_ANDROID_ARCH_ABI=arm64-v8a");
             }
 
-            if (platform.platformGroup == BuildTargetGroup.iOS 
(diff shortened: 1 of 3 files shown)


eirikwah:
Thank you for starting a PR for this. I don't have time to help right now (AVP-support in my project does not have high priority right now), however I would like to add this information:

A complete solution (not verified by me yet) is mentioned here: https://github.com/CesiumGS/cesium-unity/issues/409#issuecomment-2325620383 (pointing to https://github.com/HoloLabInc/ProjectSpirare-for-Unity/tree/main/unity/SpirareBrowser-visionOS-3DMaps#edit-cesium-for-unity and https://github.com/HoloLabInc/ProjectSpirare-for-Unity/blob/main/unity/SpirareBrowser-visionOS-3DMaps/Assets/App/Editor/CesiumPostProcessBuildForVisionOS.cs)

Here are some other links that describes some of the problems and some possible solutions:

- https://community.cesium.com/t/cesium-for-unity-on-apple-vision-pro-xcode-build-errors/32132/2
- https://community.cesium.com/t/cesium-unity-apple-vision-pro-support/32042/3
- https://community.cesium.com/t/cesium-support-for-unity-vision-pro/31010/2
- https://community.cesium.com/t/cesium-unity-on-apple-vision-pro/29862
- https://community.cesium.com/t/seams-on-google-photorealistic-tiles/30911

kring:
Thanks for the PR @handzlikchris!

> Unable to determine toolchain use for arm64-ios-unity with with CMAKE_SYSTEM_NAME visionOS. Did you mean to use VCPKG_CHAINLOAD_TOOLCHAIN_FILE?

Hmm it kind of seems like vcpkg doesn't understand visionOS as a system name. Information on this seems to be really sparse in my attempts to search so far. Worst case, it'll be necessary to define a custom toolchain.

A stupid question first, though... is visionOS really different from iOS? What I mean is, for fairly low-level code like cesium-native and its dependencies, is it possible to target a "least common denominator" of iOS and visionOS such that we don't need separate binaries for each? Or are they really such different operating systems?

handzlikchris:
@kring That's good question. To be honest I don't know how different VisionOS is from IOS. I don't have much experience with cmake, though that it may be a low-hanging fruit to change target to VisionOS for someone with more knowledge there.

If not - it may well be that building for IOS is good enough

Editor/CompileCesiumForUnityNative.cs
```diff
@@ -349,7 +377,7 @@ public static LibraryToBuild GetLibraryToBuild(PlatformToBuild platform, Library
                     library.ExtraConfigureArgs.Add("-DCMAKE_ANDROID_ARCH_ABI=arm64-v8a");
             }
 
-            if (platform.platformGroup == BuildTargetGroup.iOS)
+            if (platform.platformGroup == BuildTargetGroup.iOS || platform.platformGroup == BuildTargetGroup.VisionOS) 
             {
                 library.Toolchain = "extern/ios-toolchain.cmake";
```
Merichbier, line 382: We suppose that using same toolchain as iOS will work. However CMAKE_SYSTEM_NAME are not the same:
iOS vs visionOS
https://cmake.org/cmake/help/latest/manual/cmake-toolchains.7.html#cross-compiling-for-ios-tvos-visionos-or-watchos

raggnic:
has it been succesful for anyone? We are also currently trying to display cesium unity on vision pro

pschulzk:
@raggnic I confirm that I'm able to run https://github.com/HoloLabInc/ProjectSpirare-for-Unity/tree/main/unity/SpirareBrowser-visionOS-3DMaps (thank you, @kaorun55) with
- Unity `2022.3.20f1`
- macOS Sequoia `15.6`
- visionOS  `2.6`

However, after about 15 minutes running the app crashes. I didn't investigate but suspect memory management issues.

Thanks to @handzlikchris for this PR.
I'm very eager to see official visionOS support progressing since the industrial use cases for Cesium with visionOS are extremely promising.

Xericore:
Thanks for the quick reply. I managed to find the issue. In Unity Editor, I had to select all libraries in "Cesium for Unity/Plugins/iOS/lib" as well as "libCesiumForUnityNative" and enable the checkbox for VisionOS in "Include Platforms".

PR #503, opened 2024-09-06
Reference counting and accidental copy prevention for Impl classes

### LINKED ISSUE #223: Disable copy constructor / assignment operator on *Impl classes
CesiumGS/cesium-unity#207 was caused by accidentally copying a `CesiumIonSessionImpl` instance. This is easy to do because copying the generated C# wrappers, such as `CesiumIonSession`, is normal and safe and expected (because they act like references to C# classes), while doing the same with the Impl classes is dangerous and almost always incorrect.

We can detect and prevent bugs like this by declaring the copy constructor and assignment operator as deleted in the Impl classes. We can easily do this without a lot of boilerplate or any runtime cost by introducing a base class that our Impl classes inherit from.

kring:
### Reference Counting

Previously, C# objects backed by a native "Impl" object had strict ownership of the Impl instance. It was `new`'d in the constructor, and `delete`'d in the finalizer. This was mostly fine, but there are a few cases (mostly related to Future continuations) where it's important to keep the Impl instance alive even though the C# instance might get finalized. See #499. Previously this was not really possible because of the strict ownership. In this PR, Impl instances are internally reference counted, so their lifetime can be managed with an `IntrusivePointer`.

Note that I haven't actually done anything to take advantage of this in this PR. This PR is just the mostly-mechanical refactoring to reference counting.

### Copy prevention

This PR also makes it impossible to accidentally copy Impl classes by declaring deleted constructors and assignment operators. Accidental copying has been responsible for at least one bug in the past, and it was really simple to do this as part of the reference counting change, so I did.

Fixes #223

### DIFF
--- a/Reinterop~/MethodsImplementedInCpp.cs
+++ b/Reinterop~/MethodsImplementedInCpp.cs
@@ -33,7 +33,9 @@ namespace Reinterop
                     #endif
                     void* {{createName}}(void* handle) {
                       const {{wrapperType.GetFullyQualifiedName()}} wrapper{{{objectHandleType.GetFullyQualifiedName()}}(handle)};
-                      return reinterpret_cast<void*>(new {{implType.GetFullyQualifiedName()}}(wrapper));
+                      auto pImpl = new {{implType.GetFullyQualifiedName()}}(wrapper);
+                      pImpl->addReference();
+                      return reinterpret_cast<void*>(pImpl);
                     }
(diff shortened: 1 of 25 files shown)


j9liu:
Looks good to me @kring . Not sure what is the best way to test this, but at the very least I built and ran the branch, and connected to Cesium ion with no issues.

PR #518, opened 2024-10-15
Make cesium-unity compatible with the shared asset changes in cesium-native.

azrogers:
As we've discussed previously, we're not intending to fully implement the shared assets system in Unity at this point, as there are currently higher priority items that we need to get to. However, the changes made in cesium-native to support shared assets currently break the cesium-unity build. This change is the absolute minimum required to get cesium-unity working with the shared-assets branch of cesium-native. It doesn't implement the shared asset depot or change texture creation in any way.


j9liu:
@azrogers seems like the previous CI failures were flubs, so don't worry about them anymore. I'll give this a lookover later today.

Issue #516, opened 2024-10-14
Standardize CesiumRuntime.asmdef Platform Includes


Reag:
Currently, the CesiumRuntime.asmdef in 'Cesium For Unity/Runtime' is configured to be included on all platforms. However, Cesium for Unity does not support all platforms, and the required code gen to compile will not function. This means that if you have a special build target that does NOT involve Cesium For Unity, it will still fail compilation.

In our case, we have a headless server that manages the Unity scene's state. It performs no graphical operation and does not reference any cesium monobehaviours. We have all of our hooks into Cesium abstracted to a single asmdef that this target does not include. However, our build pipeline still fails.

The solution we would like to see is that the CesiumRuntime.asmdef is configured to only be included on platforms that Cesium for Unity supports. After all, this is why unity's Assembly Definition Asset is configured the way it is, so that you can have conditional libraries load based on the platform you are working on. Thus, if you have a Linux server that doesn't reference the CesiumRuntime, you wont fail to build (If it DOES reference the CesiumRuntime, it would fail to build, which would be the correct behavior)

kring:
That makes sense to me @Reag. Do you think you could open a pull request with that change?

Reag:
I've opened a PR with the changes that should solve this problem, found here: [PR_521](https://github.com/CesiumGS/cesium-unity/pull/521). I have also signed the CLA!

PR #521, opened 2024-10-21
Add asmdef Restrictions to Unity Runtime

Reag:
Solution to issue [https://github.com/CesiumGS/cesium-unity/issues/516](url)

* feature: Added restrictions to the CesiumRuntime.asmdef to prevent it from attempting to load on platforms that don't have a valid CesiumForUnity native plugin


azrogers:
Looks good to me! Thanks for the change, @Reag!

Issue #517, opened 2024-10-14
Cesium does not compile for Android, probably because of swl-variant


jqntn:
Tested for v1.12.0.

Here is my `build.log` grabbed from `com.cesium.unity\native~\build-Android-arm64`:

```
-- Android: Selected unified Clang toolchain
-- The CXX compiler identification is Clang 9.0.8
-- The C compiler identification is Clang 9.0.8
-- Detecting CXX compiler ABI info
-- Detecting CXX compiler ABI info - done
-- Check for working CXX compiler: N:/toolchains/llvm/prebuilt/windows-x86_64/bin/clang++.exe - skipped
-- Detecting CXX compile features
-- Detecting CXX compile features - done
-- Detecting C compiler ABI info
-- Detecting C compiler ABI info - done
-- Check for working C compiler: N:/toolchains/llvm/prebuilt/windows-x86_64/bin/clang.exe - skipped
-- Detecting C compile features
-- Detecting C compile features - done
CMake Deprecation Warning at extern/tidy-html5/CMakeLists.txt:20 (cmake_minimum_required):
  Compatibility with CMake < 3.5 will be removed from a future version of
  CMake.

  Update the VERSION argument <min> value or use a ...<max> suffix to tell
  CMake that the project does not need compatibility with older versions.


CMake Warning (dev) at extern/tidy-html5/CMakeLists.txt:26 (project):
  Policy CMP0048 is not set: project() command manages VERSION variables.
  Run "cmake --help-policy CMP0048" for policy details.  Use the cmake_policy
  command to set the policy and suppress this warning.

  The following variable(s) would be set to empty:

    PROJECT_VERSION
    PROJECT_VERSION_MAJOR
    PROJECT_VERSION_MINOR
    PROJECT_VERSION_PATCH
This warning is for project developers.  Use -Wno-dev to suppress it.

-> TIDY_SO_VERSION = 59. This is an ODD (development) release.
CMake Warning (dev) at extern/tidy-html5/CMakeLists.txt:96 (option):
  Policy CMP0077 is not set: option() honors normal variables.  Run "cmake
  --help-policy CMP0077" for policy details.  Use the cmake_policy command to
  set the policy and suppress this warning.

  For compatibility with older versions of CMake, option is clearing the
  normal variable 'BUILD_SHARED_LIB'.
This warning is for project developers.  Use -Wno-dev to suppress it.

[...]

PR #523, opened 2024-10-29
Reflect new Unity 2022.3+ requirement in developer-setup.md

jqntn:
Hi,

Regarding https://github.com/CesiumGS/cesium-unity/issues/517, here is an updated `developer-setup.md` including the correct required version for Unity.


kring:
Thanks @jqntn!

PR #524, opened 2024-10-31
Update cesium-native to latest

azrogers:
Just an oversight we made with #518, the cesium-native branch wasn't switched back to main after the shared asset changes were merged into main. This shouldn't cause any issues, but wanted to make it a PR so CI could run its checks.

### DIFF
--- a/native~/Editor/src/CesiumIonSessionImpl.cpp
+++ b/native~/Editor/src/CesiumIonSessionImpl.cpp
@@ -4,6 +4,7 @@
 #include "UnityExternals.h"
 #include "UnityTaskProcessor.h"
 
+#include <CesiumAsync/AsyncSystem.h>
 #include <CesiumUtility/Uri.h>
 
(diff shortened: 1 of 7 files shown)


azrogers:
@kring Looks like all the build issues have been fixed. Windows build on CI failed due to the same transient error I was getting on the last PR - seems like the CI machine might be running out of memory? It went away last time when we ran CI again, so hopefully it will go away again this time.

kring:
This build is failing because the UWP x86-64 native build is failing. Looking at the build logs (they're published as a build artifact), the error is this:

```
D:\cesium\CesiumForUnityBuildProject\Packages\com.cesium.unity\native~\Runtime\src\UnityPrepareRendererResources.cpp(1727,33): error C4996: 'CesiumGltfReader::GltfReader::generateMipMaps': Use ImageDecoder::generateMipMaps instead.
```

So that should be an easy fix. I'll do it.

PR #526, opened 2024-11-15
Fix automatic generation of WMTS Tile Matrix Set Labels

kring:
This was causing WMTS URLs to be built incorrectly, and so the overlay would fail to work.
Reported here:
https://community.cesium.com/t/ue5-3-unity2021/36516

### DIFF
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -6,6 +6,10 @@
 
 - Added a "From Ellipsoid" option to `Cesium3DTileset` to generate a tileset by tesselating the surface of the ellipsoid, producing a simple globe tileset without terrain features.
 
+##### Fixes :wrench:
+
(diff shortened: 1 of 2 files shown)


j9liu:
Thanks @kring ! I didn't test this -- don't think we have test data on hand for me to reproduce the error? -- but the change makes sense to me.

kring:
Yeah I don't have good test data for this either. I tested it by putting `http://localhost/{TileMatrix}` in the `url` field and `Test:` in the `tileMatrixSetLabelPrefix` field. Requests for that URL would of course fail, and Cesium for Unity would print the failed URL. Before this fix, it would print `http://localhost/`, which is clearly wrong. After this fix, it would print `http://localhost/Test:0`, which is correct.

PR #534, opened 2024-12-09
Improve Unity C# Documentation

azrogers:
In line with CesiumGS/cesium-unreal#1566 and the improvements to Cesium Native's documentation, this PR does the same for Cesium for Unity. 

- Doxygen Awesome CSS theme added, as done in Unreal and Native. Needed to be a submodule here as NPM will choke when trying to install dependencies from the Unity package.json.
- Organized existing documentation pages, including the Reinterop README, into something resembling a page hierarchy.
- Fixed the warnings reported by Doxygen (not many!)
- Changed CI to use the latest version of Doxygen, as done in Unreal and Native.

Probably the biggest thing to note is that the `Documentation~` folder had to be renamed to `Documentation`, as Doxygen seems to interpret an image at `Documentation~/images/image_name.jpg` as an image at `Documentation` with the caption `~/images/image_name.jpg`.


kring:
> Probably the biggest thing to note is that the Documentation~ folder had to be renamed to Documentation

That's probably going to cause problems, because Unity is going to pick up the folder as assets now. There are some other options for hiding assets, but none of them are great:
https://docs.unity3d.com/Manual/SpecialFolders.html

Any ideas?

azrogers:
My first idea was to [report the tilde issue on the Doxygen GitHub repo](https://github.com/doxygen/doxygen/issues/11273) but that's hardly an immediate fix. I can try out some other workarounds.

azrogers:
@kring Added the workaround we talked about, renaming Documentation back to Documentation~ and copying the images over when building the docs (using a Node script to get around cross-platform issues with `cp` vs `copy`).

PR #538, opened 2024-12-16
Fix shared image crash

### LINKED ISSUE #537: Race condition with shared images leads to crash
* Create a new scene with a Cesium3D Tileset with asset ID `2757071` from the CesiumJS Cesium ion account.
* Double-click to zoom to it
* Set the georeference origin to this location
* Move the camera so you're closer to the tileset and in front of it.
* If you haven't already crashed, click Refresh Tileset until you do.

The problem is that this tileset uses shared images, so multiple tiles refer to a single image. Each tile is loaded in a separate thread, so if two threads happen to, for example, hit `generateMipMaps` for the shared image at the same time, both will be trying to modify it simultaneously. 

Possible solutions:

* Set `GltfReaderOptions::pSharedAssetSystem` to `nullptr` so that images are never shared.
* Detect shared images while doing Unity-specific resource prep and copy them for each tile.
* Add actual support for shared images to Cesium for Unity.

The last is obviously the best by far, but may be a significant amount of work.

kring:
Fixes #537 

This is pretty much the simplest possible fix. It copies shared images so that they are no longer shared.

### DIFF
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -1,5 +1,11 @@
 # Change Log
 
+## ? - ?
+
+##### Fixes :wrench:
+
+- Fixed a bug that could cause a crash or incorrect textures when multiple `Cesium3DTileset` tiles referenced the same image by URL.
+
 ## v1.14.0 - 2024-12-02
 
 ##### Additions :tada:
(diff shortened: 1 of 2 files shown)


azrogers:
Fix looks right to me. It could be just `if (pImage->pAsset && pImage->pAsset->getDepot())` instead of the explicit null check, right?

kring:
Yeah, good point, changed. In general I don't mind being explicit about comparing to nullptr, but it was weird that I was comparing one pointer with an implicit check, and one pointer with an explicit check. I have no idea why I did that. 😆

azrogers:
With that, looks good! Thanks @kring!

PR #547, opened 2025-01-23
Update cesium-native, don't derive CesiumImpl from ReferenceCountedThreadSafe

kring:
The only change here is to `CesiumImpl`, necessitated by a change to `ReferenceCounted`.

CesiumGS/cesium-native#1048 slightly changed how we do CRTP in `ReferenceCounted`. See a bit of discussion about it here:
https://github.com/CesiumGS/cesium-native/pull/1048#discussion_r1913026176

This is a good change, but unfortunately it breaks the way we were using it in Unity. In Unity, before this PR, we were deriving `CesiumImpl` from `ReferenceCountedThreadSafe`, like this:

```cpp
template <typename TDerived>
class CesiumImpl : public CesiumUtility::ReferenceCountedThreadSafe<TDerived> {
  // ...
}
```

So essentially `CesiumImpl` is a CRTP class itself. It's essential that `TDerived` be given as the template parameter to `ReferenceCountedThreadSafe` because, when the reference count goes to zero, `ReferenceCountedThreadSafe` static casts to that type before deleting it. The destructor is not virtual (by design) so we must cast to the fully-derived type in order to properly delete the object.

Unfortunately, this doesn't work after CesiumGS/cesium-native#1048. `CesiumImpl` isn't allowed to call the private constructor in `ReferenceCountedThreadSafe` because it is not a friend; only `TDerived` is. We get compiler errors because the `ReferenceCountedThreadSafe` constructor is not accessible.

I went around in circles a bit on how to solve this. We could go back to the old CRTP approach with a protected (instead of private) constructor, and disable the clang-tidy check that got us here. Perhaps us a static_assert to achieve something close to the safety that the private constructor / friend approach was getting us (but I can't see how to get all the way there). We could add another template parameter to `ReferenceCountedThreadSafe`, so that separate classes can be specified as the "derived" and the "friend". But then couldn't that be used to subvert the safety that the private constructor is meant to provide?

In the end, I decided to leave cesium-native alone and change `CesiumImpl` to no longer derive from `ReferenceCountedThreadSafe`. This is a somewhat unusual case, so I guess it's worth a bit of duplication to get what we need here, plus the safety of the private constructor in more common scenarios.


kring:
Merging this to unbreak main.

PR #548, opened 2025-01-24
Add exception support to Reinterop

### LINKED ISSUE #18: Add support for exceptions to Reinterop

kring:
Exceptions thrown in C# can be caught in C++, and exceptions thrown in C++ can be caught in C#. Exceptions thrown in C# can also now safely pass _through_ C++ code to be code in other C# code, and vice-versa.

This is a big improvement because previously:

- C# exceptions would silently pass through C++ code without calling destructors. 💥
- C++ exceptions that escaped from the C++ code would pretty much just kill the process. 💥💥 

Fixes #18 
It works pretty much just as I described in this comment:
https://github.com/CesiumGS/cesium-unity/issues/18#issuecomment-2379146137


kring:
This still has a problem when constructors have an exception parameter in C# but not C++, so don't merge it yet.

kring:
Ok this should be ready to go now!

azrogers:
The PR looks good, but seems like there's still a build error on Windows.

azrogers:
The build error is that openssl is failing to build on ARM64 Android. I have no idea what we could've changed to make that happen. Re-running the build to see if it's just a fluke.

kring:
Yeah that intermittent Android build error has been happening for awhile. As far as I can tell, it's some sort of race condition in OpenSSL's custom build process. Running the build again usually fixes it.

PR #558, opened 2025-02-28
handle CesiumGeometry::BoundingCylinderRegion in std::variant visitor

timoore:
This won't work until, but will be necessary when, https://github.com/CesiumGS/cesium-native#1111 is included in cesium-unity's cesium-native.


kring:
Thanks @timoore!

PR #560, opened 2025-03-21
Fix crash resulting from GameObject being deleted before load thread finishes

azrogers:
A [user reported on our forums](https://community.cesium.com/t/editor-crash-issue-with-cesium-for-unity-package-1-15-4/39424) a crash in Cesium for Unity while reloading the scene repeatedly. I was able to reproduce this issue and track it down to `UntiyPrepareRendererResources::prepareInLoadThread`. When Unity toggles between play mode and edit mode, GameObjects in the scene are deleted and recreated. The problem this causes is that, if this happens right at the moment we're in the `thenInWorkerThread` part of `prepareInLoadThread`, the tileset GameObject is cleaned up by the time we return to the main thread and try to grab a component from it. All we need to do to avoid this is to check if the GameObject is valid before attempting to call `GetComponent` on it. Of course, as is the case with race conditions, I'm not positive I completely solved the issue here, but I was no longer able to reproduce the error after this fix.

### DIFF
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -1,5 +1,11 @@
 # Change Log
 
+## ? - ?
+
+##### Fixes :wrench:
+
+- Fixed a bug that could cause a crash when reloading a scene or tileset repeatedly.
+
 ## v1.15.4 - 2025-03-03
 
 ##### Fixes :wrench:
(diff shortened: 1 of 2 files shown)


kring:
Thanks @azrogers! I think this fix should be reasonably solid, because both the destruction of the tileset and the new check you've add here should happen in the main thread, which means they'll be serialized. In any case, a nullptr check certainly won't do any harm.

I updated CHANGES.md. Merging!

PR #565, opened 2025-04-16
Add `CesiumUrlTemplateRasterOverlay` component

azrogers:
This PR adds support for the UrlTemplateRasterOverlay already added to Cesium Native and Cesium for Unreal. It allows URL templates like `https://tile.openstreetmap.org/{z}/{x}/{reverseY}.png` to be used as raster overlays.

### DIFF
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -1,6 +1,12 @@
 # Change Log
 
-## v1.15.5
+## ? - ?
+
+##### Additions :tada:
+
+- Added `CesiumUrlTemplateRasterOverlay` component, allowing a raster overlay to be added using tiles requested based on a specified URL template.
+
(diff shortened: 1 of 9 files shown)


TomLeeLive:
@azrogers in Runtime/CesiumUrlTemplateRasterOverlay.cs
i think default value of _rectangleNorth = -90.0 should be 90.0

kring:
Thanks @TomLeeLive, I just noticed the same thing while reviewing this and pushed a fix for it.

kring (commented):
Looks great @azrogers! I especially like the template help incorporated into the UI. Just one small request below.

azrogers:
@kring Moved enum outside class, should be good to go!

kring:
Thanks @azrogers!

PR #567, opened 2025-04-17
Fix credits not appearing in Play Mode

j9liu:
Fixes #564. Definitely try to break this while testing and find any edge cases. Things I've tested so far:

- Switching quickly between Play and Edit modes
- Opening multiple simultaneous scene tabs
- Switching between Game and Scene views during Play Mode
- Adding and removing scenes

Issue #561, opened 2025-03-26
Crashes in `UnityPrepareRendererResources` when unloading and reloading scenes


azrogers:
[A user reported on our forums](https://community.cesium.com/t/editor-crash-issue-with-cesium-for-unity-package-1-15-4/39424) that there were crash issues in Cesium for Unity when unloading and reloading the same scene. I fixed one such issue in #560, but it turns out there's others. I've noticed crashes in [`UnityPrepareRendererResources::free`](https://github.com/CesiumGS/cesium-unity/blob/46728fe2f9bf72f22a071154c18d3263826ca27c/native~/Runtime/src/UnityPrepareRendererResources.cpp#L1723) as well as additional crashes in `prepareInLoadThread`, particularly while `populateMeshDataArray` is [running in the worker thread](https://github.com/CesiumGS/cesium-unity/blob/main/native~/Runtime/src/UnityPrepareRendererResources.cpp#L912).

These crashes seem to come down to the fact that managed resources can be destroyed by Unity while we're in the process of using them on the worker thread. Unfortunately, unlike the GameObject issue which could be solved by a simple null check, I can't think of a way to solve these issues without just giving up on running things in the worker thread entirely. We can't do a null check against a managed array, for example.

kring:
Our native code holds references to managed objects, which will keep them from being garbage collected. So, under normal circumstances, a managed object can't completely go away on us. However, there are two possible exceptions to this:
1. While the managed object reference will still be valid, the _value_ of it may become invalid. For example, Unity may destroy its own native object underlying a managed one. Comparing Unity objects against null actually checks this case via some dodgy operator overloading within Unity.
2. When the AppDomain unloads - which Unity is rather fond of doing - our managed references will be invalidated. In thoery we can subscribe to notification of AppDomain unload and handle it "somehow". More details about this here: https://github.com/CesiumGS/cesium-unity/pull/536

PR #570, opened 2025-04-30
Catch exceptions in `UnityPrepareRendererResources::free`

kring:
I noticed while testing #567 that Unity is a bit crashy when reloading AppDomains, which happens every time we enter or exit Play mode. This has always been an issue (see #561) but I think it has gotten worse with the multiple viewports changes. I don't think there is anything wrong with the multiple viewports feature itself, it just changes the timing of object destruction slightly.

As always, a "good" solution is elusive. But this PR improves the situation by adding a pokemon exception handler to `UnityPrepareRendererResources::free`. This function is on the critical path of AppDomain reloads, and oftens triggers exceptions when the managed world, now in a new AppDomain, is different from the one that the native objects from the previous AppDomain expect. By catching and ignoring these exceptions locally, rather than allowing them to propagate, we grealy reduce the chances of cascading failures.

I saw much improved stability when entering and exiting Play mode after making these changes.

### DIFF
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -9,6 +9,7 @@
 ##### Fixes :wrench:
 
 - Fixed a bug where credits would not display in the Game tab after entering Play Mode.
+- Improved stability during AppDomain reloads by catching exceptions thrown while freeing tiles.
 
 ## v1.15.5 - 2025-04-01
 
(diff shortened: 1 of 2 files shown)


kring:
@azrogers can I bug you to test this one, since you wrote #561?

azrogers:
@kring Seems to work well, haven't experienced any crashes while using it. I doubt this entirely solves our AppDomain reload issue but it's a big step forward. Thanks!

PR #575, opened 2025-05-16
Editor: Skip CesiumIonSession.Tick() when Unity is offline

Heisenberk-Karabay:
Problem
-------
If the Unity Editor launches without an Internet connection, CesiumIonServerManager.instance.currentSession.Tick() throws every frame because the DNS lookup for https://api.cesium.com fails.   The result is an endless stream of red-error messages:

    Exception: Request for `https://api.cesium.com/appData` failed:
    Cannot resolve destination host
    Exception: Failed to obtain _appData, can't resume connection

The spam clutters the Console and can even interrupt later domain reloads.

Change
------
Added a single guard clause at the top of
Editor/CesiumEditorUtility.UpdateIonSession():

    if (Application.internetReachability == NetworkReachability.NotReachable)
        return;

When Unity reports that the machine is offline the function now returns immediately, so the session is ticked again only after connectivity is restored.

### DIFF
--- a/Editor/CesiumEditorUtility.cs
+++ b/Editor/CesiumEditorUtility.cs
@@ -21,6 +21,8 @@ namespace CesiumForUnity
 
         static void UpdateIonSession()
         {
+            if (Application.internetReachability == NetworkReachability.NotReachable)
+                return;
             try
             {
                 CesiumIonServerManager.instance.currentSession.Tick();


j9liu:
Thanks for the PR @Heisenberk-Karabay! Could you please sign a [Contributor License Agreement](https://github.com/CesiumGS/community/tree/main/CLAs) (CLA) for us? We need this before we can take a look at your PR's contents. Thanks!

Heisenberk-Karabay:
Hello, signed it.

j9liu (commented):
Thanks for the PR @Heisenberk-Karabay!

I confirmed that this prevents the `Failed to connect...` logs from appearing in the Console, though it does not enable the tileset to appear in Play Mode. Is that expected for this PR?

I'm also not sure if the `Application.internetReachability` check is sufficient for all offline scenarios. The [Unity docs](https://docs.unity3d.com/6000.1/Documentation/ScriptReference/Application-internetReachability.html) say:

> Do not use this property to determine the actual connectivity. For example, the device can be connected to a hot spot, but not have the actual route to the network.

Still, the check is an improvement overall, and is something that we should consider putting in the Editor code anyway.

Heisenberk-Karabay:
No, that is not the intended behavior. The PR allowed me to see the tileset in Play mode. I’ll try to reproduce the issue with a fresh install. I’ll also look into the matter regarding `Application.internetReachability`. I’m going to investigate this ASAP and make the necessary changes. I can either submit them as a separate PR or include them in this one—whichever you prefer.

Heisenberk-Karabay:
No problem at all! I'll be eagerly waiting for the results of the tests. In the meantime, I'll go ahead and update the `CHANGES.md` file. Hope everything goes smoothly on your end!

Heisenberk-Karabay:
I've updated the `CHANGES.md` file—please let me know if anything looks off or needs to be adjusted.

kring:
@j9liu how are you testing this? I tried it out by putting my system in airplane mode shortly after launching Unity (I can't do it beforehand, because then I won't be able to check out the network license). When I do this, I do see the two errors in the top post, but each is only reported once. It's not clear to me what would cause those messages to be spammed repeatedly.

Heisenberk-Karabay:
I think the issue was on my side—I reported it a bit inaccurately. The error messages only show up once.

j9liu:
> @j9liu how are you testing this? I tried it out by putting my system in airplane mode shortly after launching Unity (I can't do it beforehand, because then I won't be able to check out the network license).

I tested this in the same way -- once I see Unity start up after it's verified our licenses, I disconnect myself from the internet. However, I was only seeing the `_appData` error once on my side.

Heisenberk-Karabay:
@j9liu @kring I hope I’m not imposing, but may I kindly ask if there have been any updates on the process?

j9liu:
Hi @Heisenberk-Karabay,

Sorry for the delay, and thanks for bumping this! Our team has been busy with the Dev Conference last week, plus the releases this week.

@azrogers previously worked on support for Cesium ion self-hosted, so I think she would be best equipped to test this. I've asked her to take a look so she'll post here with her results

Heisenberk-Karabay:
@j9liu No worries at all. @azrogers I agree that the solution you suggested is likely to be more robust. I’ll take a closer look at it in detail as soon as I have some time.

PR #580, opened 2025-06-11
279 setting the scale of a cesium3dtileset to zero causes editor crash

david-lively:
## Description

Fix for editor crash when setting any component of a 3d TileSet's transform.scale property to zero. 

## Author checklist

- ~[ ] I have updated the documentation as necessary.~

## Testing plan

Open any of the Cesium for Unity Samples scenes. Select any Cesium GameObject (CesiumGeoreference or any 3D Tileset), and set any component of the Transform.Scale Vector3 to zero. Observe that the editor does not crash.

### DIFF
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -5,6 +5,7 @@
 ##### Fixes :wrench:
 
 - Fixed broken support for `TRIANGLE_FAN` primitives in tile meshes.
+- Fixed editor crash when any 3D Tile scale component was set to zero. 
 
 ## v1.16.1 - 2025-06-02
 
--- a/native~/Runtime/src/CameraManager.cpp
+++ b/native~/Runtime/src/CameraManager.cpp
@@ -150,7 +150,6 @@ CameraManager::getAllCameras(
       0.0 == unityWorldToTileset[2].z)
     return {};
 
-
   CesiumGeoreference georeferenceComponent =
       tileset.gameObject().GetComponentInParent<CesiumGeoreference>();
   if (georeferenceComponent != nullptr) {
--- a/native~/Runtime/src/CameraManager.cpp
+++ b/native~/Runtime/src/CameraManager.cpp
@@ -121,12 +121,6 @@ void addActiveSceneCameraInEditor(
     SceneView lastActiveEditorView = SceneView::lastActiveSceneView();
     if (lastActiveEditorView != nullptr) {
       Camera editorCamera = lastActiveEditorView.camera();
-      // check for invalid scale
-      if (0.0 == unityWorldToTileset[0].x ||
-          0.0 == unityWorldToTileset[1].y ||
-          0.0 == unityWorldToTileset[2].z)
-        return;
-
       if (editorCamera != nullptr) {
         result.emplace_back(unityCameraToViewState(
           georeferenceComponent,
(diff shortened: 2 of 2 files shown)


CHANGES.md
```diff
@@ -5,6 +5,7 @@
 ##### Fixes :wrench:
 
 - Fixed broken support for `TRIANGLE_FAN` primitives in tile meshes.
+- Fixed editor crash when any 3D Tile scale component was set to zero.
```
j9liu: **NITPICK**: Prefer using the relevant class names (unless cumbersome)

```suggestion
- Fixed editor crash when any scale component on `Cesium3DTileset` was set to zero. 
```

native~/Runtime/src/CameraManager.cpp
```diff
@@ -144,6 +144,11 @@ CameraManager::getAllCameras(
   glm::dmat4 unityWorldToTileset =
       UnityTransforms::fromUnity(tileset.transform().worldToLocalMatrix());
 
+  // check for invalid scale
+  if (0.0 == unityWorldToTileset[0].x || 0.0 == unityWorldToTileset[1].y ||
+      0.0 == unityWorldToTileset[2].z)
+    return {};
```
j9liu: Hm, I'm surprised this isn't breaking in the case of 90 degree rotations around one of the axes. For instance, a matrix for a 90 degree rotation around the y-axis is:

```
[ 0, 0, 1, 0 ] 
[ 0, 1, 0, 0 ]
[ -1, 0, 0, 0 ] 
[ 0, 0, 0, 1]
```
That would fail the check. And yet, the check is passing because of floating point error:

image

That's convenient, but I'm not sure we should rely on this to successfully avoid the problem. Is there a more robust check we can do here?

david-lively: In this case, we're safe doing a direct compare with 0.0 instead of epsilon, because we're trying to avoid a div-by-zero when normalizing a vector. If the scale were set to, say, -1E-07, that's still nonzero, and normalization will succeed. I'll come up with some more test scenarios for the other case.

j9liu: Maybe my comment wasn't clear... I agree that the normalization should succeed for anything that isn't 0.0.

What I meant to suggest is... A valid matrix could be passed such that `matrix[0][0] = 0.0`, but its `scale != 0`. Like the rotation matrix in my previous comment. It just happens to work here because the matrix that Unity passes in has floating error. But theoretically, if the matrix were cleanly composed of  `0`s and `1`s, the check would fail for that matrix.

In other words, `matrix[0][0] = 0.0` does not always mean `scale.x = 0.0`, notably when rotation is involved

david-lively: Fixed in latest commit.

native~/Runtime/src/CameraManager.cpp
```diff
@@ -1,6 +1,7 @@
 #include "CameraManager.h"
 
 #include "Cesium3DTilesetImpl.h"
+#include "CesiumGeometry/Transforms.h"
```
j9liu: Prefer angle brackets when referencing cesium-native (and other third party libraries)

```suggestion
#include <CesiumGeometry/Transforms.h>
```

PR #588, opened 2025-07-23
Add support for viewing tiles in iso orientation

david-lively:
This PR adds support for Orthographic projection. When generating a `ViewState` from the Unity `Camera`, we now determine if the camera is using orthographic project and, if so, generate the correct corresponding `ViewState`. 

When `Camera.orthographic` is `true`,  the `Camera.orthographicSize` indicates half of the height of the view volume _in meters_. (See Unity's [Camera.orthographicSize](https://docs.unity3d.com/6000.1/Documentation/ScriptReference/Camera-orthographicSize.html) documentation for more detail on configuring the camera. )

In the Unity Inspector, these properties are listed under the Projection rollout. The `Size` property corresponds to the script property `Camera.orthographicSize`. 

Note that this PR modifies `Runtime/ConfigureReinterop.cs` to allow native access to `Camera::orthographic()` and `Camera::orthographicSize()`. Therefore, the Cesium Unity DLL will need to be recompiled. 

image

image

image

### DIFF
--- a/Runtime/Resources/CesiumDefaultTilesetMaterial.mat
+++ b/Runtime/Resources/CesiumDefaultTilesetMaterial.mat
@@ -23,15 +23,13 @@ Material:
   m_Name: CesiumDefaultTilesetMaterial
   m_Shader: {fileID: -6465566751694194690, guid: 407c0cff68611ac46a65eec87a5203f5,
     type: 3}
-  m_Parent: {fileID: 0}
-  m_ModifiedSerializedProperties: 0
   m_ValidKeywords:
   - _ALPHATEST_ON
   - _BUILTIN_ALPHATEST_ON
   - _BUILTIN_AlphaClip
-  m_InvalidKeywords:
   - _DISABLE_SSR_TRANSPARENT
   - _DOUBLESIDED_ON
+  m_InvalidKeywords: []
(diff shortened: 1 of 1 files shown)


kring:
Thanks @david-lively!

CHANGES.md
```diff
@@ -1,5 +1,10 @@
 # Change Log
 
+## ? - ?
+
+##### Fixes :wrench:
+- Fixed tile culling when in orthographic view.
```
j9liu, line 6: Sorry for my late comment, I'm wondering if we should position this as an addition instead? (The original issue was labelled "enhancement".)

david-lively: I'm not sure I agree. Before this PR, one could still switch the Unity camera to orthographic; it just didn't work correctly.

kring: We should put it in "Breaking Changes". Someone may have been relying on the previous behavior where orthographic cameras didn't load tiles. 

Just kidding of course! I think a reasonable case can be made for either Fixes or Additions, though.

j9liu: Putting it in Additions would let us bump up our minor version number :smile: But I don't mean to bikeshed, it's fine either way.

PR #589, opened 2025-07-23
Updated Developer Setup guide

david-lively:
This adds a note about modifying the `ConfigureReinterop.cs` files to force Unity to run Reinterop. 

On fresh clones of the package, Unity doesn't always run Reinterop. When that happens, it's necessary to modify `Editor/ConfigureReinterop.cs` and `Runtime/ConfigureReinterop.cs` so that Unity will detect a file change and then rebuild and run Reinterop.


kring:
Thanks @david-lively!

PR #594, opened 2025-08-05
Add ignore khr materials unlit option

david-lively:
## Description

This PR adds the option to override the `KHR_unlit_material` flag in a glTF 3DTileset. 

## Author checklist

- ~[ ] I have submitted a [Contributor License Agreement](https://github.com/CesiumGS/community/tree/main/CLAs) (only needed once).~
- ~[ ] I have updated the documentation as necessary.~

## Testing plan

Load a 3DTileset that uses `KHR_material_unlit` flag. Observe that it is rendered using the unlit material. Toggle the checkbox and observe that it is rendered with default lighting.


Runtime/Cesium3DTileset.cs
```diff
@@ -589,6 +589,34 @@ public bool generateSmoothNormals
...
+        /// Whether to ignore the KHR_materials_unlit extension on the glTF tiles in
+        /// this tileset, if it exists, and instead render with standard lighting and
+        /// shadows. This property will have no effect if the tileset does not have any
+        /// tiles that use this extension.
+        /// 
+        /// The KHR_materials_unlit extension is often applied to photogrammetry
+        /// tilesets because lighting and shadows are already baked into their
+        /// textures.
+        /// </remarks>
+        public bool ignoreKHRMaterialsUnlit
```
kring: This is more consistent with our practice, and also consistent with Microsoft's .NET Framework Design Guidelines. It's `Xml` not `XML`

```suggestion
        public bool ignoreKhrMaterialsUnlit
```

Runtime/Resources/CesiumDefaultTilesetMaterial.mat
```diff
@@ -222,4 +226,4 @@ MonoBehaviour:
   m_Script: {fileID: 11500000, guid: d0353a89b1f911e48b9e16bdc9f2e058, type: 3}
   m_Name: 
   m_EditorClassIdentifier: 
-  version: 5
+  version: 7
```
kring: Did you mean to commit this file? It doesn't seem necessary, and I'm worried the version change here could affect compatibility with older Unity versions.

native~/Runtime/package-lock.json
```diff
@@ -0,0 +1,244 @@
+{
```
kring: I don't think there's any reason to commit this file either.

native~/Runtime/src/UnityPrepareRendererResources.cpp
```diff
@@ -73,6 +73,7 @@
 
 #include <algorithm>
 #include <array>
+#include <iostream>
```
kring: Added for debugging? Please remove.

native~/Runtime/src/UnityPrepareRendererResources.cpp
```diff
@@ -894,22 +908,31 @@ UnityPrepareRendererResources::prepareInLoadThread(
     TileLoadResult tileLoadResult;
   };
 
+  CreateModelOptions options{};
+
   return asyncSystem
-      .runInMainThread([numberOfPrimitives]() {
+      .runInMainThread([numberOfPrimitives,
+                        tileset = this->_tilesetGameObject,
+                        &options]() {
```
kring: It's not safe to capture this by reference. By the time the main thread continuation runs, the local variable will likely be out of scope.

kring: I can think of two ways to deal with this:

1. Create the instance in the first `runInMainThread` continuation, and then pass it to the `thenInWorkerThread` continuation by returning it. The lambda will need to return a pair or a struct or something because it's already returning the result of `AllocateWritableMeshData`.
2. Use the `renderOptions` passed to this `prepareInLoadThread` method. I think this is what cesium-unreal does, because there's no "main thread" portion of `prepareInLoadThread` there, so option (1) doesn't work. If you set `TilesetOptions::rendererOptions`, whatever you supply there will magically be given to this function. Then you can `std::any_cast` it to what you know it to be (`CreateModelOptions` perhaps) and then capture in by value for the `thenInWorkerThread` lambda.

native~/Runtime/src/UnityPrepareRendererResources.cpp
```diff
@@ -1369,7 +1391,7 @@ void* UnityPrepareRendererResources::prepareInMainThread(
       static_cast<LoadThreadResult*>(pLoadThreadResult_));
 
   const System::Array1<UnityEngine::Mesh>& meshes = pLoadThreadResult->meshes;
-  const std::vector<CesiumPrimitiveInfo>& primitiveInfos =
+  std::vector<CesiumPrimitiveInfo>& primitiveInfos =
```
kring: Why remove the const here?

kring (commented):
Looks good @david-lively! A few small comments below.

Runtime/Resources/CesiumDefaultTilesetMaterial.mat
david-lively: Restored from main.

native~/Runtime/src/UnityPrepareRendererResources.cpp
david-lively: Artifact from a previous attempt at this. Fixed.

Runtime/Cesium3DTileset.cs
david-lively: Good point. Renameed this in both the native and C# code.

native~/Runtime/src/UnityPrepareRendererResources.cpp
david-lively: Thanks. I've moved the CreateModelOptions into tilesetOptions.rendererOptions.

PR #596, opened 2025-08-14
Generate face normals in pixel shader 

david-lively:
image

The [glTF 2.0 specification ](https://registry.khronos.org/glTF/specs/2.0/glTF-2.0.html) requires that flat normals are generated when they are missing from the source data. 

This PR modifies the `CesiumDefaultTilesetShader` to optionally generate face normals in the pixel shader. This is more space-efficient than the previous approach of duplicating vertices. 

This feature will be enabled when any of these conditions are true:
1. The `Cesium3DTileset.calculateFlatNormals` is `true`, 
2. The model does not have normals
3. The model is unlit, but `Cesium3DTileset.ignoreKhrMaterialsUnlit` is `true`. 

```
int32_t computeFlatNormalsPropertyID = materialProperties.getComputeFlatNormalsID();
bool computeFlatNormals = tilesetComponent.computeFlatNormals();
if (!computeFlatNormals && !primitiveInfo.hasNormals) {
  computeFlatNormals |= !primitiveInfo.isUnlit && primitive.mode != MeshPrimitive::Mode::POINTS;
}
material.SetFloat(computeFlatNormalsPropertyID, computeFlatNormals);
```

Note that any custom material that previously relied on CPU-generated flat normals should be updated to use the new feature. In `CesiumDefaultTilesetShader`, the necessary Shader Graph block is imlemented like so:

image

### DIFF
--- a/Editor/Cesium3DTilesetEditor.cs
+++ b/Editor/Cesium3DTilesetEditor.cs
@@ -36,6 +36,8 @@ namespace CesiumForUnity
         //private SerializedProperty _lodTransitionLength;
         private SerializedProperty _generateSmoothNormals;
 
+        private SerializedProperty _computeFlatNormals;
+
         private SerializedProperty _pointCloudShading;
 
         private SerializedProperty _showTilesInHierarchy;
@@ -84,6 +86,8 @@ namespace CesiumForUnity
             //    this.serializedObject.FindProperty("_lodTransitionLength");
             this._generateSmoothNormals =
                 this.serializedObject.FindProperty("_generateSmoothNormals");
+            this._computeFlatNormals =
+                this.serializedObject.FindProperty("_computeFlatNormals");
             this._ignoreKhrMaterialsUnlit = this.serializedObject.FindProperty("_ignoreKhrMaterialsUnlit");
 
             this._pointCloudShading = this.serializedObject.FindProperty("_pointCloudShading");
@@ -428,7 +432,14 @@ namespace CesiumForUnity
                 "rendered with smooth normals instead when the original glTF is missing normals.");
             EditorGUILayout.PropertyField(this._generateSmoothNormals, generateSmoothNormalsContent);
 
-            var ignoreKhrMaterialsUnlit = new GUIContent(
+            var computeFlatNormalsContent = new GUIContent("Calculate Flat Normals",
+                "When normals are missing in the glTF, caclulate flat normals in the pixel shader." +
+                "\n\n" +
+                "When using a custom material, the shader should expose the boolean `calculateFlatNormals` "+
+                " and implement this feature.");
+            EditorGUILayout.PropertyField(this._computeFlatNormals, computeFlatNormalsContent);
+
+            var ignoreKhrMaterialsUnlitContent = new GUIContent(
                 "Ignore KHR_materials_unlit",
                 "Whether to ignore the KHR_materials_unlit extension on the glTF tiles in "+
                 "this tileset, if it exists, and instead render with standard lighting and "+
@@ -439,7 +450,7 @@ namespace CesiumForUnity
                 "tilesets because lighting and shadows are already baked into their "+
                 "textures. "
             );
-            EditorGUILayout.PropertyField(this._ignoreKhrMaterialsUnlit, ignoreKhrMaterialsUnlit);
+            EditorGUILayout.PropertyField(this._ignoreKhrMaterialsUnlit, ignoreKhrMaterialsUnlitContent);
         }
 
         private void DrawPointCloudShadingProperties()
--- a/Runtime/Cesium3DTileset.cs
+++ b/Runtime/Cesium3DTileset.cs
@@ -590,6 +590,26 @@ namespace CesiumForUnity
         }
 
         [SerializeField]
+        private bool _computeFlatNormals;
+
+        /// <summary>
+        /// Whether to generate flat normals in the pixel shader when normals are missing in the glTF. 
+        /// </summary>
+        /// <remarks>
+        /// According to the glTF spec: "When normals are not specified, client
+        /// implementations should calculate flat normals." 
+        /// </remarks>
+        public bool computeFlatNormals
+        {
+            get => this._computeFlatNormals;
+            set
+            {
+                this._computeFlatNormals = value;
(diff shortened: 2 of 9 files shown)


native~/Runtime/src/UnityPrepareRendererResources.h
```diff
@@ -48,6 +48,13 @@ struct CesiumPrimitiveInfo {
    */
   bool isUnlit = false;
```
j9liu, line 49: Sorry for drive by comment; I took a glance at #595 post-merge and noticed that there is a typo -- `ignoreignoreKHRMaterialsUnlit` should be `ignoreKhrMaterialsUnlit`

CHANGES.md
```diff
@@ -5,6 +5,7 @@
 ##### Additions :tada:
 
 - Added option to ignore the `KHR_material_unlit` extension to force default lighting on tilesets. 
+- Moved flat normal generation into `CesiumDefaultTilesetShader` pixel shader. Custom lit materials should expose a boolean property `computeFlatNormals` and generate normals accordingly.
```
j9liu: Good thing to note, this could technically be a breaking change. If people created copies of `CesiumDefaultTilesetShader`, they'll be missing the material generation logic, so the appearance might be wrong. We should move this under a "Breaking Changes" category.

Runtime/Cesium3DTileset.cs
```diff
@@ -574,10 +574,9 @@ public Material opaqueMaterial
         /// Whether to generate smooth normals when normals are missing in the glTF.
         /// </summary>
         /// <remarks>
-        /// According to the glTF spec: "When normals are not specified, client
-        /// implementations should calculate flat normals." However, calculating flat
-        /// normals requires duplicating vertices. This option allows the glTFs to be rendered
-        /// with smooth normals instead when the original glTF is missing normals.
+        /// This option allows the glTFs to be rendered with smooth normals instead when the original glTF is missing normals. 
+        /// If normals are missing and this option is not enabled, flat normals should be generated
+        /// in the shader.
```
j9liu: **NITPICK** for wording

```suggestion
        /// This option allows glTFs to be rendered with smooth normals when the original glTF is missing normals. 
        /// If this option is not enabled and normals are missing, flat normals will be generated in the shader. 
```

Editor/Cesium3DTilesetEditor.cs
```diff
@@ -422,13 +422,13 @@ private void DrawRenderProperties()
                 "Generate Smooth Normals",
                 "Whether to generate smooth normals when normals are missing in the glTF." +
                 "\n\n" +
-                "According to the glTF spec: \"When normals are not specified, client " +
-                "implementations should calculate flat normals.\" However, calculating flat " +
-                "normals requires duplicating vertices. This option allows the glTFs to be " +
-                "rendered with smooth normals instead when the original glTF is missing normals.");
+                "This option allows the glTFs to be rendered with smooth normals when the " +
+                " original glTF is missing normals. If normals are missing and this option " +
+                "is not enabled, flat normals should be generated in the shader. "
```
j9liu: If you're okay with the tweaks to the comment in `Cesium3DTileset`, make sure to update this with the changes

CHANGES.md
j9liu: I might also tweak the wording here:

```suggestion
- Flat normals are now computed in `CesiumDefaultTilesetShader` to avoid duplicating vertices as tiles load. Existing copies of this shader should incorporate this logic using the boolean `computeFlatNormals` property to maintain a correct appearance.
```

native~/Runtime/src/UnityPrepareRendererResources.h
```diff
@@ -28,26 +28,33 @@ struct CreateModelOptions {
...
-   * DotNet::CesiumForUnity::Cesium3DTileset::ignoreignoreKHRMaterialsUnlit() is
+   * DotNet::CesiumForUnity::Cesium3DTileset::ignoreKhrMaterialsUnlit() is
    * set.
    */
   bool isUnlit = false;
 
+  /**
+   * @brief Whether the primitive contains normals
+   * @remarks If normals are not present amd using a lit material, they should
+   * be generated on the CPU or in the shader.
```
j9liu:
```suggestion
   * @brief Whether the primitive contains normals.
   * @remarks If normals are not present and the primitive uses a lit material,
   * they should be generated on the CPU or in the material's shader.
```

Runtime/Resources/CesiumDefaultTilesetMaterial.mat
```diff
@@ -101,7 +100,7 @@ Material:
     - _AlphaCutoffEnable: 1
     - _AlphaDstBlend: 0
     - _AlphaSrcBlend: 1
-    - _AlphaToMask: 0
+    - _AlphaToMask: 1
```
j9liu, line 103: This is the one property I'm concerned about changing here. Was this intentional? Might be worth spot checking various alpha-related cases like **CesiumPolygonRasterOverlay** or the [Alpha Blend Mode](https://github.com/KhronosGroup/glTF-Sample-Models/tree/main/2.0/AlphaBlendModeTest) sample model to verify that this doesn't break any appearances.

native~/Runtime/src/UnityPrepareRendererResources.cpp
```diff
@@ -645,17 +563,11 @@ void loadPrimitive(
             pBufferStart + colorByteOffset,
             stride,
             static_cast<size_t>(vertexCount),
-            computeFlatNormals,
+            // computeFlatNormals,
```
j9liu: Remove?

j9liu: You could actually go into `CopyVertexColors` and remove the `duplicateVertices` member, since that was only being used for flat normal generation.

Runtime/Resources/CesiumDefaultTilesetMaterial.mat
david-lively: I'm not certain how this was changed. I'll investigate.

david-lively: I see no reason for this change. I'll revert it.

PR #604, opened 2025-08-22
Fix detection of credit image loading

### LINKED ISSUE #585: Buggy credit system behavior in `04_CesiumSubscenes`
**Operating System:** Windows 11
**Unity version:** 2022.3.55.f1
**Cesium for Unity version**: 1.16.2

1. Open `04_CesiumSubscenes` in the Cesium for Unity Samples
2. Press Play. You should see the credits appear in the viewport as normal.

Image

And this is how the Data Attribution panel appears:

Image

3. Disable Cesium World Terrain, or check **Suspend Update** in its settings.
4. Enter Play mode again. Notice the absence of credits, despite them being required for Aerometrex Denver.

Image

**Note:** If you suspend updates for or disable Cesium World Terrain at runtime, it seems to work fine. It's when you enter play mode with disabled CWT that the credit system doesn't appear. When you re-enable CWT, then the credits will appear. 🤔 

This bug does not happen in Cesium for Unreal, so it seems specific to Unity.

j9liu:
## Description

This fixes buggy behavior in `CesiumCreditSystem`. The implementation is fragmented between Unity and our native code, where `CesiumCreditSystemImpl` handles credit update logic and broadcasts events to the Unity side. If the native code detects HTML images, they're passed to a Coroutine on the Unity side to load the image.

Unfortunately, there were multiple bugs that led to image-loading being inaccurately tracked, thus leading to missing credits:

- `CesiumCreditSystem` only incremented the loading images number for external images (via `UnityWebRequest`). A base64 image in process of decoding could never show up.
- `StartCoroutine` may not start right away, so `CesiumCreditSystemImpl` incorrectly assumed zero loading images `==` finished loading all images, and would broadcast the update early.

The latter is solved by explicitly checking the number of loading images every frame. This ensures that image-loading coroutines are always tracked properly so it can update the credit system UI.

## Issue number or link

Fixes #585.

## Author checklist

- [X] I have updated the documentation as necessary.

## Testing plan

See #585 for bug reproduction steps.

### DIFF
--- a/Runtime/CesiumCreditSystem.cs
+++ b/Runtime/CesiumCreditSystem.cs
@@ -267,14 +267,15 @@ namespace CesiumForUnity
             return _defaultCreditSystem;
         }
 
-        internal bool HasLoadingImages()
+        internal int GetNumberOfLoadingImages()
         {
-            return this._numLoadingImages > 0;
+            return this._numLoadingImages;
         }
 
         internal IEnumerator LoadImage(string url)
         {
             int index = this._images.Count;
+            this._numLoadingImages++;
 
             // Initialize a texture of arbitrary size as a placeholder,
             // so that when other images are loaded, their IDs align properly
@@ -297,7 +298,6 @@ namespace CesiumForUnity
             {
                 // Load an image from a URL.
                 UnityWebRequest request = UnityWebRequestTexture.GetTexture(url);
-                this._numLoadingImages++;
                 yield return request.SendWebRequest();
 
                 if (request.result == UnityWebRequest.Result.ConnectionError ||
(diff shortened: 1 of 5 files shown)


Runtime/CesiumCreditSystem.cs
```diff
@@ -314,10 +322,10 @@ internal IEnumerator LoadImage(string url)
                     UnityLifetime.Destroy(placeholderTexture);
                 }
 
-                this._numLoadingImages--;
             }
 
             texture.wrapMode = TextureWrapMode.Clamp;
+            this._numLoadingImages--;
```
david-lively, line 335: Is there any chance of this function being interrupted by an exception? If so, we might leave `this._numLoadingImages` in an invalid state.

david-lively (approved):
LGTM

PR #605, opened 2025-08-22
Prevent access on `null` `CesiumPointCloudShading`

j9liu:
## Description

This addresses a bug with point cloud tilesets that are created at runtime. Since `CesiumPointCloudShading` is a `class`, not a `struct`, it's possible for a `Cesium3DTileset` to be missing one, which is exactly what happens when it is created at runtime. This PR adds checks for a `null` reference, and also explicitly initializes `CesiumPointCloudShading` inline for good measure.

Ideally, `CesiumPointCloudShading` would just be a `struct` to avoid this, but C# 9.0 does not like `structs` with non-zero default values. :)

## Issue number or link

Fixes #604.

## Author checklist

- ~[] I have updated the documentation as necessary.~

## Testing plan

1. Start from `02_CesiumMelbourne` in the Cesium samples.
2. Create a new script called `PointCloudLoader` with the contents below.
3. Add the component to any arbitrary `GameObject`. I just tacked it onto the `Getting Started` object.
4. Press Play. Ensure that the console is not spammed with errors as the points load in.

```
using System.Collections;
using System.Collections.Generic;
using UnityEngine;
using CesiumForUnity;

public class PointCloudLoader : MonoBehaviour
{
    void Start()
    {
        var go = new GameObject("PointCloud");
        go.transform.SetParent(FindObjectOfType<CesiumGeoreference>().transform);
        Cesium3DTileset tileset = go.AddComponent<Cesium3DTileset>();
        tileset.ionAssetID = 43978;
    }
}
```

### DIFF
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -6,6 +6,10 @@
 
 - Added option to ignore the `KHR_material_unlit` extension to force default lighting on tilesets. 
 
+##### Fixes :wrench:
+
+- Fixed a bug where `CesiumPointCloudRenderer` could dereference a null `CesiumPointCloudShading` on `Cesium3DTileset`.
+
 ## v1.17.0 - 2025-08-01
 
 ##### Additions :tada:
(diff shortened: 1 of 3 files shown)


kring:
I was able to reproduce the console spam, so I'm merging this. Thanks @j9liu!

PR #609, opened 2025-09-25
Generate flat normals in shader

david-lively:
## Description

This PR adds flat normal generation in the pixel shader for tilesets that do not have normals. Before this change, either smooth per-vertex normals, or flat face normals would be generated on the CPU. The latter required duplicating vertices per face which requires a much larger vertex buffer. 
image
image

This behavior will be enabled by default when rendering a lit tileset (or unlit if the `Ignore Khr Unlit` option is enabled) if normals are not included in the tileset. 

## Author checklist

- [ ] I have submitted a [Contributor License Agreement](https://github.com/CesiumGS/community/tree/main/CLAs) (only needed once).

## Testing plan

Load a tileset that does not have normals. Disable `Generate Smooth Normals` and enable the `Ignore KHR Unlit` options. Note that the Normal material overlay contains no normals, but lighting is correct. One may also connect the normal generation output to the Base Color node on the


j9liu:
Not planning to resume this anytime soon, going to close for housekeeping.

PR #611, opened 2025-10-03
Flip textures and UVs

david-lively:
## Description

This PR modifies the texture and model import process to comply with Unity's U-right, V-up convention. 
The glTF 2.0 spec mandates that textures are anchored at UV (0,0) and the upper-left corner of the texture, with UVs increasing to the right and down. (See https://registry.khronos.org/glTF/specs/2.0/glTF-2.0.html#images .) 
Unity, however, assumes that UV (0,0) corresponds to the bottom-left corner of the image. Previously, the system did not flip either the textures OR the UVs, so it mostly worked, but has caused issues with tangent generation. 

In the UV overlay debug images below, the six columns on the left were loaded by Cesium, while the six on the right were loaded by [Unity glTFast](https://docs.unity3d.com/Packages/com.unity.cloud.gltfast@6.1/manual/installation.html). These were generated using [the same glTF model](https://github.com/KhronosGroup/glTF-Sample-Models/tree/main/2.0/NormalTangentTest):

This requires changes to `cesium-native` in https://github.com/CesiumGS/cesium-native/pull/1259 . 

Before this change:
image

After this change:
image

With textures flipped: 
image

## Author checklist

- [ ] I have submitted a [Contributor License Agreement](https://github.com/CesiumGS/community/tree/main/CLAs) (only needed once).
- [x] I have done a full self-review of my code.
- [x] I have updated `CHANGES.md` with a short summary of my change (for user-facing changes).
- [ ] I have added or updated unit tests to ensure consistent code coverage as necessary.
- [ ] I have updated the documentation as necessary.

## Testing plan

Load a tileset. Use Unity's Render Debugger to visualize the `Texcoord0` vertex attribute. Observe that UV coordinates increase in the right- and up-direction, and that textures are correctly flipped to match. 
image

PR #613, opened 2025-10-03
Import tangents

david-lively:
## Description

Previously, the glTF importer would ignore tangents in the model. This PR ensure that tangents in the vertex stream are properly imported and transferred to the relevant vertex buffer. 

(This depends on #611 so merge that first.)

## Author checklist

- [ ] I have submitted a [Contributor License Agreement](https://github.com/CesiumGS/community/tree/main/CLAs) (only needed once).

## Testing plan

Load a glTF that includes tangents, such as the [glTF Normal Tangent Mirror Test](https://github.com/KhronosGroup/glTF-Sample-Models/blob/main/2.0/NormalTangentMirrorTest/README.md).
In Unity's Render Debugger, activate the Tangent material override and verify that tangents are included. 
By default, Unity provides `(1,0,0)` for tangents if they are missing. The tangent vertex attribute override should show something other than `(1,0,0)` .

### DIFF
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -3,8 +3,7 @@
 ## ? - ?
 
 ##### Fixes :wrench:
-
-- Corrected glTF import process to flip textures and UV coordinates from glTF's U-right, V-down convention to comply with Unity's U-right, V-up coordinate system. 
+- Textures and UV coordinates from glTFs are now flipped to properly comply with Unity's U-right, V-up convention.
 - Modified model importer to include tangents when they are present in the source glTF. 
 
 ## v1.18.1 - 2025-10-01
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -5,7 +5,7 @@
 ##### Fixes :wrench:
 
 - Textures and UV coordinates from glTFs are now flipped to properly comply with Unity's U-right, V-up convention.
-- Vertex tangents from glTFs are properly imported. 
+- Vertex tangents are now properly imported when present in glTFs.
 
 ## v1.18.1 - 2025-10-01
 
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -5,6 +5,7 @@
 ##### Fixes :wrench:
 
 - Corrected glTF import process to flip textures and UV coordinates from glTF's U-right, V-down convention to comply with Unity's U-right, V-up coordinate system. 
+- Modified model importer to include tangents when they are present in the source glTF. 
 
 ## v1.18.1 - 2025-10-01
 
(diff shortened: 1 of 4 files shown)


native~/Runtime/src/UnityPrepareRendererResources.cpp
```diff
@@ -369,6 +369,15 @@ void loadPrimitive(
     computeFlatNormals = hasNormals = true;
   }
 
+  bool hasTangents = false;
+  auto tangentAcccessorIt = primitive.attributes.find("TANGENT");
```
j9liu:
```suggestion
  auto tangentAccessorIt = primitive.attributes.find("TANGENT");
```

CHANGES.md
```diff
@@ -5,6 +5,7 @@
 ##### Fixes :wrench:
 
 - Textures and UV coordinates from glTFs are now flipped to properly comply with Unity's U-right, V-up convention.
+- Vertex tangents from glTFs are properly imported.
```
j9liu: **NITPICK**

```suggestion
- Vertex tangents are now properly imported when present in glTFs.
```

native~/Runtime/src/UnityPrepareRendererResources.cpp
```diff
@@ -571,74 +589,76 @@ void loadPrimitive(
...
-    for (int64_t i = 0; i < vertexCount; ++i) {
-      *reinterpret_cast<Vector3*>(pWritePos) = positionView[i];
+  }
+
+  for (int64_t i = 0; i < vertexCount; ++i) {
+    const TIndex vertexIndex = duplicateVertices ? indices[i] : i;
+    *reinterpret_cast<Vector3*>(pWritePos) = positionView[vertexIndex];
+    pWritePos += sizeof(Vector3);
+
+    if (computeFlatNormals) {
```
j9liu, line 630: Hm I'm not sure how much of an impact it has here, but we tend to avoid putting `if` statements in a loop body if possible. I know in Unreal we prefer to use near-identical `for` loops if it can avoid a few `if` statements mid-loop. Maybe others have an opinion?

j9liu (commented):
Thanks @david-lively! Still waiting on my slow machine to build so I can actually try this out, but I'll leave some comments in the meantime.

native~/Runtime/src/UnityPrepareRendererResources.cpp
david-lively: @j9liu  Hmm. Not sure I see the benefit here. The existing implementation already had `if` for vertex colors and normals. To completely get rid of the conditionals, we'd need 8 nearly-identical permutations of this loop.

j9liu:
Not planning to resume this anytime soon, going to close for housekeeping.

PR #620, opened 2025-10-20
Add support for building Cesium For Unity Web

brendan-duncan:
## Description

Add support for building Cesium for Unity Web. This will build cesium-native and all dependencies with the Emscripten that is included with Unity, as part of the build process similar to building for other native platforms.

This will require Unity 6.0+ to build for the web as it requires support for C/C++ Native Multithreading. Previous versions of Unity did not have this option.
In the Unity Project Settings, select Player, Web Platform, Publishing Settings, and turn on the Enable C/C++ Native Multithreading option.

This currently only works building the static libs in the Plugins/WebGL on Windows. The resulting libraries can be used by other platforms.

The ninja executable must be found the PATH. It would be nice if this were automatically handled by CMake or installed with vcpkg but I don't know how to do that.

Some of the changes needed to get things to build are:
* If the path to Unity's Emscripten has a space in it, use subst to map the Emscripten directory to M: drive.
* blend2d is added as a submodule and built locally due to the evcpkg version causing build errors.

This PR is dependent on the changes in https://github.com/CesiumGS/cesium-native/pull/1267.

If you follow the directions to build locally in https://github.com/CesiumGS/cesium-unity/blob/main/Documentation~/developer-setup.md, if the cesium-native PR hasn't been merged yet, replace the checkout of cesium-native in native~/extern with a checkout of https://github.com/brendan-duncan/cesium-native/tree/experiment_wasm.

### DIFF
--- a/Editor/CompileCesiumForUnityNative.cs
+++ b/Editor/CompileCesiumForUnityNative.cs
@@ -648,10 +648,10 @@ namespace CesiumForUnity
                     // Refresh the asset database for platforms that use static linking so the Unity
                     // builder can find the libraries.
                     if (library.Platform == BuildTarget.iOS || library.Platform == BuildTarget.WebGL)
+                    {
                         AssetDatabase.Refresh();
-
-                    if (library.Platform == BuildTarget.WebGL)
                         SetStaticLibrariesPlatform(library);
+                    }
                 }
             }
             finally


kring:
Merging! Thanks @brendan-duncan!

TomLeeLive:
@brendan-duncan Thanks again! I understand! this PR is what I always wanted :)

Lostalie:
> The Unity Build&Run local server adds the server headers, which is why it works when testing from that. If you're hosting on your own server, what I do is I include a file called .htaccess in the same directory as the index.html, with the contents
> 
> ```
>  <FilesMatch "\.(htm|html|js|js.gz|js.br)$">
>      Header add Cross-Origin-Opener-Policy "same-origin"
>      Header add Cross-Origin-Embedder-Policy "require-corp"
>      Header add Cross-Origin-Resource-Policy "cross-origin"
>  </FilesMatch>
> ```

I create a .htaccess with contents and same directory as index html, then hosting on my Nginx Server, but it doesn't seem to work.What need do more?

PR #624, opened 2025-10-29
Attempt to work around warning spam from Unity physics engine

kring:
> Detected one or more triangles where the distance between any 2 vertices is greater than 500 units. The resulting Triangle Mesh can impact simulation and query stability. It is recommended to tessellate meshes that have large triangles.

This commit tries to avoid this warning by scaling down the vertex positions of every tile mesh to fit inside a cube with a diagonal <= 1 unit. And then scale it back up to its proper size via the transformation matrix.

It reduces the number of occurrences of this warning, but for some reason it doesn't eliminate them entirely.

Draft because I'd like to eliminate these warnings entirely, and I don't know why it isn't working.

This PR also gives each mesh a name derived from the tile's content URL, so that the warning tells us which tile is the problem. This should probably be removed before this is merged because the URLs can get quite long.

### DIFF
--- a/native~/Runtime/src/UnityPrepareRendererResources.cpp
+++ b/native~/Runtime/src/UnityPrepareRendererResources.cpp
@@ -586,6 +586,9 @@ void loadPrimitive(
   }
   stride += numTexCoords * sizeof(Vector2);
 
+  Vector3 minPosition;
+  Vector3 maxPosition;
+
   if (computeFlatNormals) {
     ::computeFlatNormals(
         pWritePos + normalByteOffset,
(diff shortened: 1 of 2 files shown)


leohilbert:
@brendan-duncan Any news on this? I'm experimenting with Cesium right now and get bombarded with these error messages (like other 1000 in 1 minute). Any way I can work around it myself?

brendan-duncan:
In the project settings under Physics, you can set the verbosity level to Default or Minimal to quiet those messages and get some much needed silence.

PR #628, opened 2025-11-14
Fix two Reinterop problems with delegates

### LINKED ISSUE #627: Native callbacks can be freed while they're running
`UnityTaskProcessor::startTask` has a one line implementation:

```cpp
void UnityTaskProcessor::startTask(std::function<void()> f) {
  System::Threading::Tasks::Task::Run(f);
}
```

`Task::Run` takes a `DotNet::System::Action` instance, which is implicitly constructed from the `std::function`.  The implicit constructor implementation looks like this:

```cpp
void* (*Action::CreateDelegate)(void* pCallbackFunction) = nullptr;

Action::Action(std::function<FunctionSignature> callback) :
    _handle(CreateDelegate(reinterpret_cast<void*>(new std::function<FunctionSignature>(std::move(callback)))))
{
}
```

So a new `std::function` is allocated on the heap and then passed to `CreateDelegate`, which is a managed function. The managed function creates an instance of a managed class called `ActionNativeFunction`. `ActionNativeFunction` holds that `std::function` pointer (as an `IntPtr`). Its `Invoke` method calls back into native code to invoke the function. Finally, that `Invoke` is hooked up to a regular managed `System.Action` delegate.

End result: the C# System.Action "owns" a `std::function` that it can call at will. That `System.Action` can be passed to, in this case, `Task.Run`, which can now effectively call back into native code.

Here's the problem. When the managed `ActionNativeFunction` is finalized, the `std::function` will be deleted. This can happen even while the `std::function` is still running! I don't know how common this is, but I have a crash up on my screen right now where this is exactly what has happened (during finalization for AppDomain unload). This is #17 again, which was mostly fixed in #177, but apparently not for delegates.

kring:
This PR fixes two problems with the generated code for delegates.

1. When a native `std::function` is wrapped with a .NET delegate, and the delegate is invoked, that native code could end up throwing an exception. Previously, this exception would have propagated disastrously through the managed code. With this PR, it's caught at the interface layer and turned into a managed exception. Just like it would be (even before this PR) if it were a partial method implemented in native code rather than a delegate.
2. Delegate wrappers for native functions weren't using `SafeHandle` to reference the native function, which meant the delegate could be garbage collected, finalized, and destroy the `std::function` while it was running. As described in #627.

Fixes #627 

Sadly this does not fix all the AppDomain unload stability problems. It's still pretty easy to get a crash by entering play mode while tilesets are loading. I think I have a pretty good idea what to do next on that front, though, so I will write it up as a separate issue.

### DIFF
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -1,5 +1,12 @@
 # Change Log
 
+## ? - ?
+
+##### Fixes :wrench:
+
+- Exceptions thrown by delegates implemented in native code are now correctly propagated back to the managed caller. Previously, this scenario could cause crashes and undefined behavior.
+- Delegate wrappers for native functions now use `SafeHandle` to ensure the native function will not be destroyed by the garbage collector while it is running.
+
 ## v1.19.0 - 2025-11-03
 
 ##### Additions :tada:


Reinterop~/CustomDelegateGenerator.cs
```diff
@@ -72,7 +72,7 @@ private void GenerateDelegate(CppGenerationContext context, GeneratedResult resu
                     """
                 ));
 
-            // A a C# delegate type that wraps a std::function, and arrange for
+            // A C# delegate type that wraps a std::function, and arrange for
```
j9liu: Typo?

```suggestion
            // A C# delegate type that wraps a std::function, and arranges for
```

Reinterop~/CustomDelegateGenerator.cs
j9liu, line 1: Is it worth noting this change in CHANGES.md?

PR #634, opened 2025-11-28
Fixed cross compilation from Linux

myss:
Example: build happens from Linux CI server, with another platform, like WebGL or Mac as target. CesiumForUnity package is installed.

The build happens by means of the Unity Editor. When the editor is opened, it does two things:

* loads the project and ensures that the editor loads the needed resources/scripts, compiles the needed scripts
* switches to the target platform to make the build, starts the build, etc

Cesium can't be compiled on Linux: the native runtime is missing for Linux, and Reinterop does not generate any Linux specific code. However, when loading the Unity Editor on Linux, it tries to compile Cesium scripts for Linux, which fail because of missing Reinterop bindings. When that fails, the cross compilation to WebGL does not start at all.

Cesium .asmdef files don't declare explicit support for Linux, which is ok. However, they declare support for "Editor", which also matches the Editor on Linux. With this commit, we ensure that Cesium scripts won't be loaded for the Editor on Linux either. This way, projects can be loaded on Linux and can be built successfully for another target platform.

There is one final problem with this approach, though: when Cesium scripts are not loaded on Linux, Cesium components added the the scene will lose their data in the Editor. These components won't be there when the project is built with another target. This should not be a problem if the Cesium components will be added to the scene at runtime. A more proper fix would probably involve to make Reinterop compilation work on Linux (with stub implementations even if Cesium native is not available for Linux).

## Author checklist

- [ ] I have updated `CHANGES.md` with a short summary of my change (for user-facing changes).


j9liu (approved):
Thanks for the PR @myss ! This change seems fair, I'm just running it through CI before we merge.

PR #639, opened 2025-12-03
Prevent OnScreenCredits from making GUI elements below non-pickable

myss:
Prevent OnScreenCredits from making GUI elements below non-pickable.

Especially on mobile (verified on Android + Google PR3DT), these credits could cover 20% of screen height at the bottom. Some of our applicaiton buttons appear there from time to time, and the credits make them un-clickable.

With this change, the credits don't prevent the GUI elements below from handling mouse actions.
## Author checklist

- [X] I have submitted a [Contributor License Agreement](https://github.com/CesiumGS/community/tree/main/CLAs) (only needed once).
- [ ] I have updated `CHANGES.md` with a short summary of my change (for user-facing changes).


myss:
Hi @j9liu .

I think both improvements are good to have:
* show credits in a smaller size
* even then, if they cover something, they should not make them unclickable

Even on desktop they overlap partially with some of our buttons.

PR #641, opened 2025-12-05
Improve performance in Debug builds

kring:
Debug builds of Cesium for Unity are so slow that they're nearly unusable. In a typical scene in the Editor with Google Photorealistic 3D Tiles, the frame rate drops below 1 FPS. Debug builds are always slow, but this is ridiculous. And it seems to be worse than it used to be.

This appears to be largely a result of MSVC's "iterator debugging" features, which are enabled by default Debug builds. For reasons that are not entirely clear to me, every time an iterator is destroyed, the iterator debugging features internally acquire some mutex. We do this a lot. Maybe more than we used to, which is why things are slower now. But iterator destruction is supposed to be cheap!

You can read more about the iterator debugging features here:
https://learn.microsoft.com/en-us/cpp/standard-library/debug-iterator-support?view=msvc-170

It might be nice to have these things enabled, but it's more important that we have usable debug builds. And hey, Unreal turns this stuff off entirely! So this PR switches to `_ITERATOR_DEBUG_LEVEL=1`. It still provides some safety, but is less horribly slow.

I also noticed while profiling that Cesium3DTileset's Update method is sometimes called more than once per frame in the Editor (_outside_ play mode). This PR fixes that, too, which will benefit even Release builds.

### DIFF
--- a/native~/vcpkg/triplets/x64-windows-unity.cmake
+++ b/native~/vcpkg/triplets/x64-windows-unity.cmake
@@ -7,3 +7,4 @@ set(VCPKG_LIBRARY_LINKAGE static)
 # In Debug builds on Windows, the default _ITERATOR_DEBUG_LEVEL=2 makes things _really_ slow,
 # for little benefit. Use a more reasonable debug level.
 set(VCPKG_CXX_FLAGS_DEBUG "/D_ITERATOR_DEBUG_LEVEL=1")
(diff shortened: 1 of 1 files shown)


j9liu (approved):
Thanks @kring ! I didn't do any formal benchmarks, but just from observation this is performing much better.

PR #642, opened 2025-12-11
Combine Runtime and Editor assemblies

kring:
Wait! Don't close this PR! Sure, it sounds like a terrible idea from the title, but bear with me... 🐻

Currently, before this PR, we have two managed (AKA .NET) assemblies: CesiumRuntime.dll and CesiumEditor.dll. CesiumEditor contains code that is only ever needed in the Editor, while CesiumRuntime contains... uh.. both runtime and Editor code. Hmm... Well, ok, that's not totally crazy. It's common for a class that is used at runtime to need just a little bit of Editor-specific code for one reason or another. Nothing wrong with that!

Ok, so we have a bit of Editor code in our CesiumRuntime assembly, for good reasons. Does that mean the CesiumRuntime assembly is a bit more bloated than it needs to be? No, not at all! The Unity compilation model is funny. Rather than build these two assemblies once and ship them off to end-user systems (which is, arguably, a big reason to used managed code in the first place!), Unity builds them separately in every unique context. So our CesiumRuntime is built once for the Editor on Windows, and again for the Player on Windows (i.e., a built game, no Editor involved). And then again for macOS Editor and again for macOS Player. And again for Android Player (no Editor here). And so on...  Ifdef's ensure each platform's build gets exactly what it needs and nothing more.

All of which is a long-winded way of explaining why the separation of our code into CesiumRuntime and CesiumEditor is completely pointless, even though it seems like a good idea on the surface. The Windows Editor is always going to load its own specific builds of both CesiumRuntime and CesiumEditor. So why separate them?

Well, there is one reason: in the case where an entire file is only suitable for use in the Editor, we can just put it in an Editor specific folder and we're done. No need to ifdef-out the whole file for the Editor. So that's kind of nice.

But what's the cost? It turns out it's quite high, because of Reinterop and because of runtime libraries. Let's talk about runtime libraries first.

When building applications for Windows, you have to [choose](https://learn.microsoft.com/en-us/cpp/build/reference/md-mt-ld-use-run-time-library?view=msvc-170) between the "DLL" runtime library (`/MD`) or the "static" runtime library (`/MT`). If you use the DLL version, then end-users of your application will need to have the MSVC [redistributable](https://learn.microsoft.com/en-us/cpp/windows/latest-supported-vc-redist?view=msvc-170) installed. And it has to be the same or later version as the one paired with the compiler you used to compiler your application, or else things can go very poorly (crashes).

Unlike Unreal Engine, Unity doesn't install an MSVC redistributable (I'm talking about the engines themselves here, not our plugins). So we have no guarantees about what version will exist on end-user systems, or we may even find that none exists at all. Forcing our developer users to install one on their end-user systems, when Unity itself does not require this, would not be very nice. So to avoid this, we use the static version of the runtime library.

Using the static runtime library means, in turn, that every DLL we build gets its own separate heap. Allocating an object with `new` in one DLL and then `delete`'ing it in another leads to undefined behavior. That means we have to be extraordinarily careful about what we pass between our DLLs. Certainly STL containers are right out. Really, we'd have to be so careful that this really isn't practical to do at all.

Which brings us to Reinterop. Reinterop is a plugin for the Roslyn C# compiler that runs while it is compiling the code in the CesiumRuntime and CesiumEditor assemblies. Because these two assemblies are compiled separately, Reinterop generates completely separate code for a corresponding native DLL for each. Changing this would be pretty challenging. The managed CesiumEditor assembly is allowed to call into the managed CesiumRuntime assembly. But because we're using the static version of the MSVC runtime library, it's not safe for the native DLL paired with CesiumEditor to call into the native DLL paired with CesiumRuntime.

This is why we currently have a `native~/Shared` folder with a bunch of code that is duplicated between the two native DLLs. And why each has its very own AsyncSystem. And why the network request cancelation implemented in #632 does not also cancel Editor network requests, such as the request to refresh the list of Cesium ion assets, potentially leading to some lingering instability on AppDomain reload.

I set out to fix that last point, and this PR is the result. My conclusion was that the least painful approach is to combine CesiumRuntime and CesiumEditor into a single managed assembly, paired with a single Reinterop native DLL.

This required a ton of moving stuff around:

* Instead of a `CesiumRuntime.dll` built from the `Runtime` directory and a `CesiumEditor.dll` built from the `Editor` directory, there's now just a top level `Source` directory that builds to `CesiumForUnity.dll`.
* The top-level `Source` directory has `Runtime` and `Editor` subdirectories, but these are just for organization. They could all be flattened into one directory and that wouldn't change anything.
* `#if UNITY_EDITOR` has been added to the top of every file in the `Editor` directory. If this were a normal csproj build system, we could probably make this happen automatically. But Unity asmdefs aren't as sophisticated, so we have to do it manually.
* Similarly, `native~` now has a `src` directory with `Editor` and `Runtime` subdirectories. Why `src` here and `Source` for the managed stuff? Just managed versus native conventions, I guess.
* The `native~/src/Editor` is automatically excluded if `UNITY_EDITOR` is not defined, because that's easy to do in our native code's cmake build system.

So this is probably the simplest 427 file PR that you'll ever review. But there are a lot of moving parts that all have to line up perfectly, and it took me a lot of iterations on CI to get it right.

I've tested the Editor on Windows and macOS, and built games on Windows, macOS, Android, and Web. I've also successfully generated an Xcode project for iOS but can't currently build it for unrelated reasons. I didn't try UWP. I don't expect any problems, but should probably try it anyway.

### DIFF
--- a/Source/Runtime/ConfigureReinterop.cs
+++ b/Source/Runtime/ConfigureReinterop.cs
@@ -41,7 +41,7 @@ namespace CesiumForUnity
 #else
         public const string CppOutputPath = "../../native~/generated-Unknown";
 #endif
-//
+
         // The namespace with which to prefix all C# namespaces. For example, if this
         // property is set to "DotNet", then anything in the "System" namespace in C#
         // will be found in the "DotNet::System" namespace in C++.


Editor/NativeDownloadHandler.cs
```diff
@@ -1,35 +0,0 @@
...
-using System;
-using UnityEngine.Networking;
-
-namespace CesiumForUnity
-{
-    public struct RawDownloadedData
-    {
-        public IntPtr pointer;
-        public int length;
-    }
```
j9liu, line 11: Just curious why this was needed for the Editor version of the file but not `Runtime/NativeDownloadHandler.cs`?

.gitignore
```diff
@@ -18,10 +18,12 @@ Reinterop.deps.json
...
-libCesiumForUnityNative-*.dylib.meta
+CesiumForUnityNative.dll
+CesiumForUnityNative.dll.meta
+libCesiumForUnityNative.so
+libCesiumForUnityNative.so.meta
+libCesiumForUnityNative.dylib
+libCesiumForUnityNative.dylib.meta
 Documentation~/Reference/
+generated~/
+test-server~/
```
j9liu: How might these files get generated? I don't see them on my end, at least.

Source/Runtime/ConfigureReinterop.cs
```diff
@@ -27,21 +27,21 @@ internal partial class ConfigureReinterop
...
+        public const string CppOutputPath = "../../native~/generated-Standalone";
 #elif UNITY_WEBGL
-        public const string CppOutputPath = "../native~/Runtime/generated-WebGL";
+        public const string CppOutputPath = "../../native~/generated-WebGL";
 #else
-        public const string CppOutputPath = "../native~/Runtime/generated-Unknown";
+        public const string CppOutputPath = "../../native~/generated-Unknown";
 #endif
-
+//
```
j9liu: **NIT**: leftover characters?

native~/package-lock.json
```diff
@@ -0,0 +1,233 @@
+{
```
j9liu, line 1: Silly question but what is this `package-lock.json` doing? It wasn't needed before, was it?

Build~/Package.cs
```diff
@@ -426,56 +432,64 @@ private static void CopyPackageContents(string sourcePath, string targetPath)
...
+                "EditorTests.meta",
                 "ThirdParty.json",
                 "ThirdParty.json.meta"
             };
 
             foreach (string file in filesToCopy)
             {
-                File.Copy(Path.Combine(sourcePath, file), Path.Combine(targetPath, file));
+                string completeSource = Path.Combine(sourcePath, file);
+                if (File.Exists(completeSource))
```
j9liu, line 455: These `Exists` checks weren't there before; are they needed for any particular reason now? Or should they have always been there for safety?

Editor/NativeDownloadHandler.cs
kring: It wasn't needed in either one. I don't know why it existed (unused) in the Editor version. Probably just hanging around from a previous attempt or approach.

.gitignore
kring: I often use `generated~` as the Reinterop output directory, i.e., when compiling with `-generatedfilesout` (see [here](https://github.com/CesiumGS/cesium-unity/blob/main/Reinterop~/README.md#temporarily-disabling-the-code-generator)) to look at the generated code.

`test-server~` was where I stored a node.js server I used to test Unity on the web.

No good reason I should have committed either, though. I'll remove them.

native~/package-lock.json
kring: Not strictly needed, perhaps, but it's a good idea. It ensures that `npm install` in the `native~` directory gets the exact versions of the npm packages. So we don't get a new version of clang-format unexpectedly, for example.

Build~/Package.cs
kring: They're needed because some of the paths might not exist now. Because the `Editor` directory is now only created when we do an Editor build, and because we don't do an Editor build at all in the Emscripten job on CI, that directory won't exist. Without the check, the build would fail.

The others probably aren't strictly needed, but getting to this stage and then failing a build because a file is (correctly) missing is really annoying. Ask me how I know.

kring:
I've seen the "invalid GC Handle" error a couple of times now! Unfortunately there's no call stack associated with it, which is going to make it especially fun to track down.

j9liu:
Whoops @kring, I realized I forgot to ask whether it was worth mentioning this to `CHANGES.md`! But we can just commit to main if so.

PR #656, opened 2026-01-27
Vector overlay

baruchInsert-tech:
## Author checklist

### DIFF
--- a/native~/src/Runtime/CesiumGeoJsonDocumentImpl.cpp
+++ b/native~/src/Runtime/CesiumGeoJsonDocumentImpl.cpp
@@ -1,4 +1,5 @@
 #include "CesiumGeoJsonDocumentImpl.h"
+
 #include "CesiumGeoJsonObjectImpl.h"
 #include "UnityTilesetExternals.h"
 
@@ -59,8 +60,8 @@ bool CesiumGeoJsonDocumentImpl::ParseInternal(
     System::String geoJsonString) {
   std::string str = geoJsonString.ToStlString();
 
-  Result<GeoJsonDocument> result = GeoJsonDocument::fromGeoJson(
-      std::span<const std::byte>(
+  Result<GeoJsonDocument> result =
+      GeoJsonDocument::fromGeoJson(std::span<const std::byte>(
           reinterpret_cast<const std::byte*>(str.data()),
           str.size()));
 
@@ -81,18 +82,16 @@ void CesiumGeoJsonDocumentImpl::LoadFromUrl(
   std::shared_ptr<IAssetAccessor> pAssetAccessor = getAssetAccessor();
 
   GeoJsonDocument::fromUrl(asyncSystem, pAssetAccessor, urlStr)
-      .thenInMainThread(
-          [callback](Result<GeoJsonDocument>&& result) {
-            if (!result.value.has_value()) {
-              callback.Invoke(CesiumForUnity::CesiumGeoJsonDocument(nullptr));
-              return;
-            }
+      .thenInMainThread([callback](Result<GeoJsonDocument>&& result) {
+        if (!result.value.has_value()) {
+          callback.Invoke(CesiumForUnity::CesiumGeoJsonDocument(nullptr));
+          return;
+        }
 
-            CesiumForUnity::CesiumGeoJsonDocument doc;
-            doc.NativeImplementation().setNativeDocument(
-                std::move(*result.value));
-            callback.Invoke(doc);
-          });
+        CesiumForUnity::CesiumGeoJsonDocument doc;
+        doc.NativeImplementation().setNativeDocument(std::move(*result.value));
+        callback.Invoke(doc);
+      });
 }
 
 void CesiumGeoJsonDocumentImpl::LoadFromCesiumIon(
@@ -112,18 +111,16 @@ void CesiumGeoJsonDocumentImpl::LoadFromCesiumIon(
       ionAssetId,
       accessToken,
       apiUrl)
-      .thenInMainThread(
-          [callback](Result<GeoJsonDocument>&& result) {
-            if (!result.value.has_value()) {
-              callback.Invoke(CesiumForUnity::CesiumGeoJsonDocument(nullptr));
-              return;
-            }
+      .thenInMainThread([callback](Result<GeoJsonDocument>&& result) {
+        if (!result.value.has_value()) {
+          callback.Invoke(CesiumForUnity::CesiumGeoJsonDocument(nullptr));
+          return;
+        }
 
-            CesiumForUnity::CesiumGeoJsonDocument doc;
-            doc.NativeImplementation().setNativeDocument(
-                std::move(*result.value));
-            callback.Invoke(doc);
-          });
+        CesiumForUnity::CesiumGeoJsonDocument doc;
+        doc.NativeImplementation().setNativeDocument(std::move(*result.value));
+        callback.Invoke(doc);
+      });
 }
 
 } // namespace CesiumForUnityNative
--- a/native~/src/Runtime/CesiumGeoJsonDocumentRasterOverlayImpl.cpp
+++ b/native~/src/Runtime/CesiumGeoJsonDocumentRasterOverlayImpl.cpp
@@ -6,22 +6,20 @@
 #include "CesiumVectorStyleConversions.h"
 #include "UnityExternals.h"
 
-#include <DotNet/CesiumForUnity/CesiumGeoJsonDocument.h>
-
 #include <Cesium3DTilesSelection/Tileset.h>
 #include <CesiumRasterOverlays/GeoJsonDocumentRasterOverlay.h>
 #include <CesiumUtility/Color.h>
 #include <CesiumVectorData/GeoJsonDocument.h>
 #include <CesiumVectorData/VectorStyle.h>
 
-#include <spdlog/spdlog.h>
-
 #include <DotNet/CesiumForUnity/Cesium3DTileset.h>
+#include <DotNet/CesiumForUnity/CesiumGeoJsonDocument.h>
 #include <DotNet/CesiumForUnity/CesiumGeoJsonDocumentRasterOverlay.h>
 #include <DotNet/CesiumForUnity/CesiumGeoJsonDocumentRasterOverlaySource.h>
 #include <DotNet/CesiumForUnity/CesiumIonServer.h>
 #include <DotNet/CesiumForUnity/CesiumRasterOverlay.h>
 #include <DotNet/System/String.h>
+#include <spdlog/spdlog.h>
 
 using namespace Cesium3DTilesSelection;
 using namespace CesiumRasterOverlays;
@@ -32,15 +30,14 @@ namespace {
 
 CesiumAsync::Future<std::shared_ptr<CesiumVectorData::GeoJsonDocument>>
 wrapLoaderFuture(
-    CesiumAsync::Future<CesiumUtility::Result<CesiumVectorData::GeoJsonDocument>>&&
-        future) {
+    CesiumAsync::Future<
+        CesiumUtility::Result<CesiumVectorData::GeoJsonDocument>>&& future) {
   return std::move(future).thenImmediately(
       [](CesiumUtility::Result<CesiumVectorData::GeoJsonDocument>&&
              documentResult)
           -> std::shared_ptr<CesiumVectorData::GeoJsonDocument> {
         if (documentResult.errors) {
-          spdlog::default_logger()->error(
-              "GeoJSON document has errors!");
+          spdlog::default_logger()->error("GeoJSON document has errors!");
           documentResult.errors.logError(
               spdlog::default_logger(),
               "Errors loading GeoJSON document: ");
@@ -88,7 +85,8 @@ void CesiumGeoJsonDocumentRasterOverlayImpl::AddToTileset(
 
   const CesiumGeospatial::Ellipsoid& ellipsoid = pTileset->getEllipsoid();
 
-  CesiumVectorData::VectorStyle nativeStyle = CesiumForUnityNative::fromUnityStyle(overlay.defaultStyle());
+  CesiumVectorData::VectorStyle nativeStyle =
+      CesiumForUnityNative::fromUnityStyle(overlay.defaultStyle());
 
   GeoJsonDocumentRasterOverlayOptions vectorOptions{
       nativeStyle,
@@ -126,9 +124,9 @@ void CesiumGeoJsonDocumentRasterOverlayImpl::AddToTileset(
           vectorOptions,
           options);
     }
-  } else if (source ==
-             CesiumForUnity::CesiumGeoJsonDocumentRasterOverlaySource::
-                 FromDocument) {
+  } else if (
+      source ==
+      CesiumForUnity::CesiumGeoJsonDocumentRasterOverlaySource::FromDocument) {
     // FromDocument - use a pre-parsed and styled document
     CesiumForUnity::CesiumGeoJsonDocument doc = overlay.document();
     if (doc == nullptr) {
--- a/native~/src/Runtime/CesiumGeoJsonFeatureImpl.cpp
+++ b/native~/src/Runtime/CesiumGeoJsonFeatureImpl.cpp
@@ -1,4 +1,5 @@
 #include "CesiumGeoJsonFeatureImpl.h"
+
 #include "CesiumGeoJsonObjectImpl.h"
 
 #include <CesiumUtility/JsonValue.h>
@@ -10,7 +11,6 @@
 #include <DotNet/CesiumForUnity/CesiumGeoJsonFeatureIdType.h>
 #include <DotNet/CesiumForUnity/CesiumGeoJsonObject.h>
 #include <DotNet/System/String.h>
-
 #include <rapidjson/document.h>
 #include <rapidjson/stringbuffer.h>
 #include <rapidjson/writer.h>
--- a/native~/src/Runtime/CesiumGeoJsonObjectImpl.cpp
+++ b/native~/src/Runtime/CesiumGeoJsonObjectImpl.cpp
@@ -1,4 +1,5 @@
 #include "CesiumGeoJsonObjectImpl.h"
+
 #include "CesiumGeoJsonFeatureImpl.h"
 #include "CesiumVectorStyleConversions.h"
 
@@ -69,7 +70,9 @@ CesiumGeoJsonObjectImpl::GetObjectAsFeature(
   }
 
   CesiumForUnity::CesiumGeoJsonFeature result;
-  result.NativeImplementation().setNativeFeatureInDocument(_pDocument, pFeature);
+  result.NativeImplementation().setNativeFeatureInDocument(
+      _pDocument,
+      pFeature);
   return result;
 }
 
@@ -96,7 +99,9 @@ CesiumGeoJsonObjectImpl::GetObjectAsFeatureCollection(
 
     CesiumForUnity::CesiumGeoJsonFeature feature;
     if (pFeature) {
-      feature.NativeImplementation().setNativeFeatureInDocument(_pDocument, pFeature);
+      feature.NativeImplementation().setNativeFeatureInDocument(
+          _pDocument,
+          pFeature);
     }
     result.Item(i, feature);
   }
@@ -162,8 +167,7 @@ CesiumGeoJsonObjectImpl::GetObjectAsLineString(
   System::Array1<Unity::Mathematics::double3> points(count);
 
   for (std::int32_t i = 0; i < count; ++i) {
-    const glm::dvec3& coord =
-        pLineString->coordinates[static_cast<size_t>(i)];
+    const glm::dvec3& coord = pLineString->coordinates[static_cast<size_t>(i)];
     points.Item(i, Unity::Mathematics::double3{coord.x, coord.y, coord.z});
   }
 
@@ -267,16 +271,13 @@ CesiumGeoJsonObjectImpl::GetObjectAsMultiPolygon(
     System::Array1<CesiumForUnity::CesiumGeoJsonLineString> rings(ringCount);
 
     for (std::int32_t r = 0; r < ringCount; ++r) {
-      const std::vector<glm::dvec3>& ring =
-          polyRings[static_cast<size_t>(r)];
+      const std::vector<glm::dvec3>& ring = polyRings[static_cast<size_t>(r)];
       std::int32_t pointCount = static_cast<std::int32_t>(ring.size());
 
       System::Array1<Unity::Mathematics::double3> points(pointCount);
       for (std::int32_t j = 0; j < pointCount; ++j) {
         const glm::dvec3& coord = ring[static_cast<size_t>(j)];
-        points.Item(
-            j,
-            Unity::Mathematics::double3{coord.x, coord.y, coord.z});
+        points.Item(j, Unity::Mathematics::double3{coord.x, coord.y, coord.z});
       }
 
       CesiumForUnity::CesiumGeoJsonLineString lineString;
--- a/native~/src/Runtime/CesiumGeoJsonObjectImpl.h
+++ b/native~/src/Runtime/CesiumGeoJsonObjectImpl.h
@@ -33,13 +33,14 @@ public:
       const DotNet::CesiumForUnity::CesiumGeoJsonObject& object);
   ~CesiumGeoJsonObjectImpl();
 
-  // Set an object that references data within a document (no copy, modifications persist)
+  // Set an object that references data within a document (no copy,
+  // modifications persist)
   void setNativeObjectInDocument(
       std::shared_ptr<CesiumVectorData::GeoJsonDocument> pDocument,
       CesiumVectorData::GeoJsonObject* pObject);
 
-  std::int32_t GetObjectType(
-      const DotNet::CesiumForUnity::CesiumGeoJsonObject& object);
+  std::int32_t
+  GetObjectType(const DotNet::CesiumForUnity::CesiumGeoJsonObject& object);
 
   bool IsValid(const DotNet::CesiumForUnity::CesiumGeoJsonObject& object);
 
@@ -86,9 +87,7 @@ public:
 
   void ClearStyle(const DotNet::CesiumForUnity::CesiumGeoJsonObject& object);
 
-  CesiumVectorData::GeoJsonObject* getNativeObject() const {
-    return _pObject;
-  }
+  CesiumVectorData::GeoJsonObject* getNativeObject() const { return _pObject; }
 
   std::shared_ptr<CesiumVectorData::GeoJsonDocument> getDocument() const {
     return _pDocument;
--- a/Source/Editor/CesiumGeoJsonDocumentRasterOverlayEditor.cs.meta
+++ b/Source/Editor/CesiumGeoJsonDocumentRasterOverlayEditor.cs.meta
@@ -1,0 +1,2 @@
+fileFormatVersion: 2
+guid: f2c7d3f3ee4c2ed46a9e447bd94a3c01
--- a/Source/Runtime/CesiumGeoJsonDocument.cs.meta
+++ b/Source/Runtime/CesiumGeoJsonDocument.cs.meta
@@ -1,0 +1,2 @@
+fileFormatVersion: 2
+guid: d99548f43a931a14f97b63535fa2db47
--- a/Source/Runtime/CesiumGeoJsonDocumentRasterOverlay.cs.meta
+++ b/Source/Runtime/CesiumGeoJsonDocumentRasterOverlay.cs.meta
@@ -1,0 +1,2 @@
+fileFormatVersion: 2
+guid: 28822e61bf1dd21488cbba1ec740739e
--- a/Source/Runtime/CesiumGeoJsonFeature.cs.meta
+++ b/Source/Runtime/CesiumGeoJsonFeature.cs.meta
@@ -1,0 +1,2 @@
+fileFormatVersion: 2
+guid: 43a3c6536d69ef148b8c8f44c429bd84
--- a/Source/Runtime/CesiumGeoJsonFeatureIdType.cs.meta
+++ b/Source/Runtime/CesiumGeoJsonFeatureIdType.cs.meta
@@ -1,0 +1,2 @@
+fileFormatVersion: 2
+guid: 1a887d54bac04b94a8408a139f369269
--- a/Source/Runtime/CesiumGeoJsonLineString.cs.meta
+++ b/Source/Runtime/CesiumGeoJsonLineString.cs.meta
@@ -1,0 +1,2 @@
+fileFormatVersion: 2
+guid: 5ec2fcedd33fc0e41b8711b1ca3fbd45
--- a/Source/Runtime/CesiumGeoJsonObject.cs.meta
+++ b/Source/Runtime/CesiumGeoJsonObject.cs.meta
@@ -1,0 +1,2 @@
+fileFormatVersion: 2
+guid: 2c8e83f81ad7637469b9201ba061ce5c
--- a/Source/Runtime/CesiumGeoJsonObjectType.cs.meta
+++ b/Source/Runtime/CesiumGeoJsonObjectType.cs.meta
@@ -1,0 +1,2 @@
+fileFormatVersion: 2
+guid: 027d86ab5d61ccd4d8bfa20189c1fffd
--- a/Source/Runtime/CesiumGeoJsonPolygon.cs.meta
+++ b/Source/Runtime/CesiumGeoJsonPolygon.cs.meta
@@ -1,0 +1,2 @@
+fileFormatVersion: 2
+guid: 4180e57135ea1cf48b9aa7c92fba8563
--- a/Source/Runtime/CesiumVectorStyle.cs.meta
+++ b/Source/Runtime/CesiumVectorStyle.cs.meta
@@ -1,0 +1,2 @@
+fileFormatVersion: 2
+guid: dce3577060bd32a45907a151c5422eb2
--- a/Tests/TestCesiumGeoJsonDocument.cs.meta
+++ b/Tests/TestCesiumGeoJsonDocument.cs.meta
@@ -1,0 +1,2 @@
+fileFormatVersion: 2
+guid: 07f69c7a5ac1b8740b2918d448fee2aa
--- a/Tests/TestCesiumGeoJsonObject.cs.meta
+++ b/Tests/TestCesiumGeoJsonObject.cs.meta
@@ -1,0 +1,2 @@
+fileFormatVersion: 2
+guid: 86e44aba1153e0f4b9ba10c95e2c9f5a
--- a/Tests/TestCesiumVectorStyle.cs.meta
+++ b/Tests/TestCesiumVectorStyle.cs.meta
@@ -1,0 +1,2 @@
+fileFormatVersion: 2
+guid: 2683dc9d613dd5149b9eab2384e74d55
--- a/native~/src/Runtime/CesiumGeoJsonDocumentRasterOverlayImpl.h
+++ b/native~/src/Runtime/CesiumGeoJsonDocumentRasterOverlayImpl.h
@@ -19,7 +19,8 @@ class CesiumGeoJsonDocumentRasterOverlayImpl
     : public CesiumImpl<CesiumGeoJsonDocumentRasterOverlayImpl> {
 public:
   CesiumGeoJsonDocumentRasterOverlayImpl(
-      const DotNet::CesiumForUnity::CesiumGeoJsonDocumentRasterOverlay& overlay);
+      const DotNet::CesiumForUnity::CesiumGeoJsonDocumentRasterOverlay&
+          overlay);
   ~CesiumGeoJsonDocumentRasterOverlayImpl();
 
   void AddToTileset(
--- a/native~/src/Runtime/CesiumGeoJsonFeatureImpl.h
+++ b/native~/src/Runtime/CesiumGeoJsonFeatureImpl.h
@@ -16,8 +16,7 @@ class CesiumGeoJsonObject;
 
 namespace CesiumForUnityNative {
 
-class CesiumGeoJsonFeatureImpl
-    : public CesiumImpl<CesiumGeoJsonFeatureImpl> {
+class CesiumGeoJsonFeatureImpl : public CesiumImpl<CesiumGeoJsonFeatureImpl> {
 public:
   CesiumGeoJsonFeatureImpl(
       const DotNet::CesiumForUnity::CesiumGeoJsonFeature& feature);
--- a/native~/src/Runtime/CesiumVectorStyleConversions.h
+++ b/native~/src/Runtime/CesiumVectorStyleConversions.h
@@ -3,13 +3,13 @@
 #include <CesiumUtility/Color.h>
 #include <CesiumVectorData/VectorStyle.h>
 
-#include <DotNet/UnityEngine/Color32.h>
 #include <DotNet/CesiumForUnity/CesiumVectorColorMode.h>
 #include <DotNet/CesiumForUnity/CesiumVectorLineStyle.h>
 #include <DotNet/CesiumForUnity/CesiumVectorLineWidthMode.h>
 #include <DotNet/CesiumForUnity/CesiumVectorPolygonFillStyle.h>
 #include <DotNet/CesiumForUnity/CesiumVectorPolygonStyle.h>
 #include <DotNet/CesiumForUnity/CesiumVectorStyle.h>
+#include <DotNet/UnityEngine/Color32.h>
 
 #include <cstdint>
 
@@ -34,8 +34,8 @@ fromUnityLineWidthMode(DotNet::CesiumForUnity::CesiumVectorLineWidthMode mode) {
       static_cast<std::uint8_t>(mode));
 }
 
-inline CesiumVectorData::LineStyle
-fromUnityLineStyle(const DotNet::CesiumForUnity::CesiumVectorLineStyle& lineStyle) {
+inline CesiumVectorData::LineStyle fromUnityLineStyle(
+    const DotNet::CesiumForUnity::CesiumVectorLineStyle& lineStyle) {
   CesiumVectorData::LineStyle native;
   native.color = fromUnityColor(lineStyle.color);
   native.colorMode = fromUnityColorMode(lineStyle.colorMode);
@@ -44,8 +44,8 @@ fromUnityLineStyle(const DotNet::CesiumForUnity::CesiumVectorLineStyle& lineStyl
   return native;
 }
 
-inline CesiumVectorData::ColorStyle
-fromUnityFillStyle(const DotNet::CesiumForUnity::CesiumVectorPolygonFillStyle& fillStyle) {
+inline CesiumVectorData::ColorStyle fromUnityFillStyle(
+    const DotNet::CesiumForUnity::CesiumVectorPolygonFillStyle& fillStyle) {
   CesiumVectorData::ColorStyle native;
   native.color = fromUnityColor(fillStyle.color);
   native.colorMode = fromUnityColorMode(fillStyle.colorMode);
@@ -62,7 +62,8 @@ fromUnityStyle(const DotNet::CesiumForUnity::CesiumVectorStyle& style) {
   }
 
   if (style.polygonStyle.outline) {
-    native.polygon.outline = fromUnityLineStyle(style.polygonStyle.outlineStyle);
+    native.polygon.outline =
+        fromUnityLineStyle(style.polygonStyle.outlineStyle);
   }
 
   return native;


baruchInsert-tech:
> Thank you so much for the contribution @baruchInsert-tech! Just to confirm, would you like a review on this PR? I see that it's marked "Draft", so we weren't sure if it was still in progress.

Hello!
this PR is ready for review.
Thank you!

asafMasa:
> Thank you so much for the contribution @baruchInsert-tech! Just to confirm, would you like a review on this PR? I see that it's marked "Draft", so we weren't sure if it was still in progress.

Hi @j9liu , is there any update on this PR review? 
Do you have an estimation when it can be reviewed?
It solves a long standing issue in unity - https://community.cesium.com/t/display-geojson-data-on-map/27125
Thanks.

Source/Runtime/CesiumGeoJsonDocumentRasterOverlay.cs
```diff
@@ -0,0 +1,236 @@
...
+        /// <summary>
+        /// The raster overlay will load a GeoJSON document from Cesium ion.
+        /// </summary>
+        FromCesiumIon = 1,
+
+        /// <summary>
+        /// The raster overlay will use a GeoJSON document that has been parsed
+        /// and styled in code using <see cref="CesiumGeoJsonDocumentRasterOverlay.SetGeoJsonDocument"/>.
+        /// </summary>
+        FromDocument = 2
```
j9liu: Cesium for Unreal uses `FromDocument = 0` and `FromUrl = 2` for this enum: https://github.com/CesiumGS/cesium-unreal/blob/main/Source/CesiumRuntime/Public/CesiumGeoJsonDocumentRasterOverlay.h#L18-L32

Could you change this here for consistency?

Source/Runtime/CesiumGeoJsonDocumentRasterOverlay.cs
```diff
@@ -0,0 +1,236 @@
...
+    [ReinteropNativeImplementation(
+        "CesiumForUnityNative::CesiumGeoJsonDocumentRasterOverlayImpl",
+        "CesiumGeoJsonDocumentRasterOverlayImpl.h")]
+    [AddComponentMenu("Cesium/Cesium GeoJSON Document Raster Overlay")]
+    [IconAttribute("Packages/com.cesium.unity/Editor/Resources/Cesium-24x24.png")]
+    public partial class CesiumGeoJsonDocumentRasterOverlay : CesiumRasterOverlay
+    {
+        [SerializeField]
+        private CesiumGeoJsonDocumentRasterOverlaySource _source =
+            CesiumGeoJsonDocumentRasterOverlaySource.FromUrl;
```
j9liu: Cesium ion is our default source option in Unreal: https://github.com/CesiumGS/cesium-unreal/blob/main/Source/CesiumRuntime/Public/CesiumGeoJsonDocumentRasterOverlay.h#L50

```suggestion
            CesiumGeoJsonDocumentRasterOverlaySource.FromCesiumIon;
```

Source/Runtime/CesiumGeoJsonDocumentRasterOverlay.cs
```diff
@@ -0,0 +1,236 @@
...
+        /// You can style individual features in the document before calling this method
+        /// by using <see cref="CesiumGeoJsonObject.SetStyle"/>.
+        /// </remarks>
+        /// <param name="geoJsonDocument">The parsed GeoJSON document to use.</param>
+        public void SetGeoJsonDocument(CesiumGeoJsonDocument geoJsonDocument)
+        {
+            this._document = geoJsonDocument;
+            this._source = CesiumGeoJsonDocumentRasterOverlaySource.FromDocument;
+            this.Refresh();
+        }
```
j9liu: Is there a reason that this doesn't follow the `get`/`set` conventions of the other parameters?

Source/Runtime/CesiumGeoJsonDocumentRasterOverlay.cs
```diff
@@ -0,0 +1,236 @@
...
+
+        /// <summary>
+        /// Internal method to get the document for native code.
+        /// </summary>
+        internal CesiumGeoJsonDocument GetDocumentInternal()
+        {
+            return this._document;
+        }
```
j9liu: Is this truly necessary when `public CesiumGeoJsonDocument document` exists earlier in the document?

Source/Runtime/CesiumGeoJsonDocumentRasterOverlay.cs
```diff
@@ -0,0 +1,236 @@
...
+
+        /// <inheritdoc/>
+        protected override partial void RemoveFromTileset(Cesium3DTileset tileset);
+
+#if UNITY_EDITOR
+        private void Reset()
+        {
+            this._defaultStyle = CesiumVectorStyle.Default;
+        }
+#endif
```
j9liu: What is this used for? Seems unused in this PR.

native~/src/Runtime/CesiumGeoJsonDocumentRasterOverlayImpl.cpp
```diff
@@ -0,0 +1,288 @@
...
+        vectorOptions,
+        options);
+  } else if (source ==
+             CesiumForUnity::CesiumGeoJsonDocumentRasterOverlaySource::
+                 FromDocument) {
+    // FromDocument - use a pre-parsed and styled document
+    CesiumForUnity::CesiumGeoJsonDocument doc = overlay.GetDocumentInternal();
+    if (doc == nullptr) {
+      spdlog::default_logger()->error(
+          "GeoJSON document is null! Call SetGeoJsonDocument first.");
```
j9liu: It's okay to just exit early. Otherwise the user gets noise in the console when they're adjusting the overlay settings.

```suggestion
```

native~/src/Runtime/CesiumGeoJsonDocumentRasterOverlayImpl.cpp
```diff
@@ -0,0 +1,288 @@
...
+    this->_pOverlay = new GeoJsonDocumentRasterOverlay(
+        overlay.materialKey().ToStlString(),
+        getAsyncSystem().createResolvedFuture(std::move(pDocument)),
+        vectorOptions,
+        options);
+  } else {
+    // FromUrl
+    System::String url = overlay.url();
+    if (System::String::IsNullOrEmpty(url)) {
+      spdlog::default_logger()->error("GeoJSON URL is empty!");
```
j9liu: It's okay to just exit early. Otherwise the user gets noise in the console when they're adjusting the overlay settings.

```suggestion
```

native~/src/Runtime/CesiumGeoJsonDocumentRasterOverlayImpl.cpp
```diff
@@ -0,0 +1,288 @@
...
+
+  if (nativeStyle.polygon.fill.has_value()) {
+    spdlog::default_logger()->info(
+        "GeoJSON polygon fill color: ({}, {}, {}, {})",
+        nativeStyle.polygon.fill->color.r,
+        nativeStyle.polygon.fill->color.g,
+        nativeStyle.polygon.fill->color.b,
+        nativeStyle.polygon.fill->color.a);
+  }
+
```
j9liu: Were these used for debugging purposes? We don't log this in Unreal and it would be nice to remove to reduce noise.

native~/src/Runtime/CesiumGeoJsonDocumentRasterOverlayImpl.cpp
```diff
@@ -0,0 +1,288 @@
...
+CesiumGeoJsonDocumentRasterOverlayImpl::
+    ~CesiumGeoJsonDocumentRasterOverlayImpl() {}
+
+void CesiumGeoJsonDocumentRasterOverlayImpl::AddToTileset(
+    const ::DotNet::CesiumForUnity::CesiumGeoJsonDocumentRasterOverlay& overlay,
+    const ::DotNet::CesiumForUnity::Cesium3DTileset& tileset) {
+  spdlog::default_logger()->info("CesiumGeoJsonDocumentRasterOverlay::AddToTileset called");
+
+  if (this->_pOverlay != nullptr) {
+    spdlog::default_logger()->info("Overlay already added, returning");
```
j9liu:
```suggestion
```

native~/src/Runtime/CesiumGeoJsonDocumentRasterOverlayImpl.cpp
```diff
@@ -0,0 +1,288 @@
...
+    return;
+  }
+
+  Cesium3DTilesetImpl& tilesetImpl = tileset.NativeImplementation();
+  Tileset* pTileset = tilesetImpl.getTileset();
+  if (!pTileset) {
+    spdlog::default_logger()->error("No tileset found!");
```
j9liu:
```suggestion
```

native~/src/Runtime/CesiumGeoJsonDocumentRasterOverlayImpl.cpp
```diff
@@ -0,0 +1,288 @@
...
+    this->_pOverlay = new GeoJsonDocumentRasterOverlay(
+        overlay.materialKey().ToStlString(),
+        wrapLoaderFuture(GeoJsonDocument::fromCesiumIonAsset(
+            getAsyncSystem(),
+            getAssetAccessor(),
+            overlay.ionAssetID(),
+            ionAccessToken.ToStlString(),
+            apiUrl)),
+        vectorOptions,
+        options);
```
j9liu: This should be encased in an `if (!apiUrl.empty()) { ... }`, in case `apiUrl` is empty.

https://github.com/CesiumGS/cesium-unity/blob/fe0f617803cfb43a6ccb83436763964afb150df1/native~/src/Runtime/CesiumIonRasterOverlayImpl.cpp#L62-L75

native~/src/Runtime/CesiumGeoJsonDocumentRasterOverlayImpl.cpp
```diff
@@ -0,0 +1,288 @@
...
+  }
+
+  spdlog::default_logger()->info(
+      "GeoJSON overlay created with materialKey: {}",
+      overlay.materialKey().ToStlString());
+
+  pTileset->getOverlays().add(this->_pOverlay);
+  spdlog::default_logger()->info("GeoJSON overlay added to tileset");
```
j9liu: Remove the logs here too

native~/src/Runtime/CesiumGeoJsonDocumentRasterOverlayImpl.cpp
```diff
@@ -0,0 +1,288 @@
...
+  } else {
+    // FromUrl
+    System::String url = overlay.url();
+    if (System::String::IsNullOrEmpty(url)) {
+      spdlog::default_logger()->error("GeoJSON URL is empty!");
+      return;
+    }
+
+    std::string urlStr = url.ToStlString();
+    spdlog::default_logger()->info("Loading GeoJSON from URL: {}", urlStr);
```
j9liu:
```suggestion
```

native~/src/Runtime/CesiumGeoJsonDocumentRasterOverlayImpl.cpp
```diff
@@ -0,0 +1,288 @@
...
+    // Use the shared_ptr directly to preserve per-feature styles
+    std::shared_ptr<GeoJsonDocument> pDocument = docImpl.getSharedDocument();
+    if (!pDocument) {
+      spdlog::default_logger()->error(
+          "GeoJSON document shared_ptr is null!");
+      return;
+    }
+
+    spdlog::default_logger()->info(
+        "Loading GeoJSON from pre-parsed document with per-feature styles");
```
j9liu:
```suggestion
      return;
    }
```

Source/Editor/CesiumGeoJsonDocumentRasterOverlayEditor.cs
```diff
@@ -0,0 +1,153 @@
...
+        private void DrawStyleProperties()
+        {
+            EditorGUILayout.LabelField("Style", EditorStyles.boldLabel);
+
+            GUIContent mipLevelsContent = new GUIContent(
+                "Mip Levels",
+                "The number of mip levels to generate for each tile of this raster overlay. " +
+                "Additional mip levels can improve the visual quality of tiles farther from " +
+                "the camera at the cost of additional rasterization time.");
+            EditorGUILayout.IntSlider(this._mipLevels, 0, 8, mipLevelsContent);
```
j9liu: I would make this a `DelayedIntField`; we usually don't use sliders

native~/src/Runtime/CesiumGeoJsonDocumentImpl.cpp
```diff
@@ -0,0 +1,138 @@
...
+                std::move(*result.value));
+            callback.Invoke(doc);
+          });
+}
+
+void CesiumGeoJsonDocumentImpl::DisposeNative(
+    const CesiumForUnity::CesiumGeoJsonDocument& document) {
+  _pDocument = nullptr;
+  _isValid = false;
+}
```
j9liu: Is there a reason that this isn't done in `~CesiumGeoJsonDocumentImpl`? We'll use `Dispose` methods for careful lifecycle management purposes, but this doesn't seem particularly unwieldy.

native~/src/Runtime/CesiumGeoJsonDocumentImpl.cpp
```diff
@@ -0,0 +1,138 @@
...
+bool CesiumGeoJsonDocumentImpl::ParseInternal(
+    const CesiumForUnity::CesiumGeoJsonDocument& document,
+    System::String geoJsonString) {
+  std::string str = geoJsonString.ToStlString();
+
+  std::vector<std::byte> bytes(str.size());
+  std::memcpy(bytes.data(), str.data(), str.size());
+
+  Result<GeoJsonDocument> result =
+      GeoJsonDocument::fromGeoJson(std::span<const std::byte>(bytes));
```
j9liu: Could we perhaps avoid this `memcpy` by creating a `std::span<const std::byte>` on `str.data()`?

native~/src/Runtime/CesiumGeoJsonDocumentImpl.cpp
```diff
@@ -0,0 +1,138 @@
...
+CesiumGeoJsonDocumentImpl::~CesiumGeoJsonDocumentImpl() {}
+
+void CesiumGeoJsonDocumentImpl::setNativeDocument(GeoJsonDocument&& document) {
+  _pDocument = std::make_shared<GeoJsonDocument>(std::move(document));
+  _isValid = true;
+}
+
+bool CesiumGeoJsonDocumentImpl::IsValid(
+    const CesiumForUnity::CesiumGeoJsonDocument& document) {
+  return _isValid && _pDocument != nullptr;
```
j9liu: `_isValid` isn't doing too much throughout this file because we're always checking for `pDocument != nullptr`. I would just remove it.

native~/src/Runtime/CesiumGeoJsonDocumentImpl.h
```diff
@@ -0,0 +1,65 @@
...
+      DotNet::System::String ionAccessToken,
+      DotNet::System::String ionApiUrl,
+      DotNet::System::Action1<DotNet::CesiumForUnity::CesiumGeoJsonDocument>
+          callback);
+
+  void DisposeNative(const DotNet::CesiumForUnity::CesiumGeoJsonDocument& document);
+
+  const CesiumVectorData::GeoJsonDocument& getNativeDocument() const {
+    return *_pDocument;
+  }
```
j9liu: This doesn't seem to be used, and it's redundant with `getSharedDocument` anyways. Remove?

Source/Runtime/CesiumGeoJsonDocument.cs
```diff
@@ -0,0 +1,133 @@
...
+        /// Internal method to parse GeoJSON string into this document.
+        /// </summary>
+        internal partial bool ParseInternal(string geoJsonString);
+
+        /// <summary>
+        /// Loads a GeoJSON document from a URL asynchronously.
+        /// </summary>
+        /// <param name="url">The URL to load the GeoJSON from.</param>
+        /// <returns>A task that resolves to the loaded CesiumGeoJsonDocument, or null if loading failed.</returns>
+        public static async Task<CesiumGeoJsonDocument> LoadFromUrlAsync(string url)
```
j9liu, line 65: Do you have an example of how to use this async action? I've only had experience with Unity Coroutines, so I'm a little unfamiliar with using `async`/`await` directly. :o

Source/Runtime/CesiumGeoJsonDocument.cs
```diff
@@ -0,0 +1,133 @@
...
+        /// </summary>
+        internal partial void DisposeNative();
+
+        /// <summary>
+        /// Finalizer to release native resources.
+        /// </summary>
+        ~CesiumGeoJsonDocument()
+        {
+            DisposeNative();
+        }
```
j9liu: See my comment in `CesiumGeoJsonDocumentImpl.cpp`; I'm not sure this is necessary, but if there's a good reason to do this we'll want to know.

native~/src/Runtime/CesiumGeoJsonDocumentRasterOverlayImpl.cpp
```diff
@@ -0,0 +1,288 @@
...
+CesiumAsync::Future<std::shared_ptr<CesiumVectorData::GeoJsonDocument>>
+wrapLoaderFuture(
+    CesiumAsync::Future<CesiumUtility::Result<CesiumVectorData::GeoJsonDocument>>&&
+        future) {
+  return std::move(future).thenImmediately(
+      [](CesiumUtility::Result<CesiumVectorData::GeoJsonDocument>&&
+             documentResult)
+          -> std::shared_ptr<CesiumVectorData::GeoJsonDocument> {
+        spdlog::default_logger()->info(
+            "GeoJSON document loader future resolved");
```
j9liu:
```suggestion
```

native~/src/Runtime/CesiumGeoJsonDocumentRasterOverlayImpl.cpp
```diff
@@ -0,0 +1,288 @@
...
+        }
+
+        if (!documentResult.value.has_value()) {
+          spdlog::default_logger()->error(
+              "GeoJSON document result has no value!");
+          return nullptr;
+        }
+
+        spdlog::default_logger()->info(
+            "GeoJSON document loaded successfully");
```
j9liu:
```suggestion
```

Source/Runtime/CesiumGeoJsonObject.cs
```diff
@@ -0,0 +1,167 @@
...
+    /// Represents an object in a GeoJSON document.
+    /// </summary>
+    /// <remarks>
+    /// A GeoJSON object can be a geometry (Point, LineString, Polygon, etc.),
+    /// a Feature (geometry with properties), or a collection of other objects.
+    /// </remarks>
+    [ReinteropNativeImplementation(
+        "CesiumForUnityNative::CesiumGeoJsonObjectImpl",
+        "CesiumGeoJsonObjectImpl.h")]
+    public partial class CesiumGeoJsonObject
```
j9liu, line 20: Can you walk me through the decisions you made for the API here? It differs a bit from the Cesium for Unreal API, which isn't inherently bad, but it'd be good to understand the reasons why.

In [Cesium for Unreal](https://github.com/CesiumGS/cesium-unreal/blob/main/Source/CesiumRuntime/Public/CesiumGeoJsonObject.h) we have different GeoJSON object subtypes, like `FCesiumGeoJsonFeature`, `FCesiumGeoJsonLineString`, and so forth. But here everything is merged into one API.

Source/Runtime/CesiumGeoJsonDocumentRasterOverlay.cs
baruchInsert-tech: You're right, removed it. the native code now calls overlay.document() directly through the property getter.

Source/Runtime/CesiumGeoJsonDocumentRasterOverlay.cs
baruchInsert-tech: Refactored SetGeoJsonDocument into a standard get/set property to follow the conventions of the other parameters. the setter still automatically switches source to FromDocument, since setting a document implies the intent to use it.

Source/Runtime/CesiumGeoJsonDocumentRasterOverlay.cs
baruchInsert-tech: Changed

Source/Runtime/CesiumGeoJsonDocumentRasterOverlay.cs
baruchInsert-tech: Removed. It was only resetting _defaultStyle to the same value as the field initializer, so it wasn't doing anything    meaningful.

Source/Runtime/CesiumGeoJsonDocumentRasterOverlay.cs
baruchInsert-tech: Changed to match Unreal's enum values: FromDocument = 0, FromCesiumIon = 1, FromUrl = 2.

native~/src/Runtime/CesiumGeoJsonDocumentRasterOverlayImpl.cpp
baruchInsert-tech: Removed

native~/src/Runtime/CesiumGeoJsonDocumentRasterOverlayImpl.cpp
baruchInsert-tech: Removed

native~/src/Runtime/CesiumGeoJsonDocumentRasterOverlayImpl.cpp
baruchInsert-tech: Removed

native~/src/Runtime/CesiumGeoJsonDocumentRasterOverlayImpl.cpp
baruchInsert-tech: Removed

native~/src/Runtime/CesiumGeoJsonDocumentRasterOverlayImpl.cpp
baruchInsert-tech: Done

native~/src/Runtime/CesiumGeoJsonDocumentRasterOverlayImpl.cpp
baruchInsert-tech: Done

native~/src/Runtime/CesiumGeoJsonDocumentRasterOverlayImpl.cpp
baruchInsert-tech: Removed

native~/src/Runtime/CesiumGeoJsonDocumentRasterOverlayImpl.cpp
baruchInsert-tech: Removed

Source/Editor/CesiumGeoJsonDocumentRasterOverlayEditor.cs
baruchInsert-tech: Done

native~/src/Runtime/CesiumGeoJsonDocumentImpl.cpp
baruchInsert-tech: You're right, moved the cleanup into ~CesiumGeoJsonDocumentImpl and removed DisposeNative entirely, the C# finalizer is no longer needed.

native~/src/Runtime/CesiumGeoJsonDocumentImpl.cpp
baruchInsert-tech: Replaced the memcpy and vector with a std::span directly on str.data(), and cleaned up the unused includes

native~/src/Runtime/CesiumGeoJsonDocumentImpl.cpp
baruchInsert-tech: Done

native~/src/Runtime/CesiumGeoJsonDocumentImpl.h
baruchInsert-tech: Done

Source/Runtime/CesiumGeoJsonDocument.cs
baruchInsert-tech: In Unity, you can use async/await directly in MonoBehaviour methods. For example:                                     
```
async void Start()                                                                                                      
{
      var document = await CesiumGeoJsonDocument.LoadFromUrlAsync("https://example.com/data.geojson");
      if (document != null && document.IsValid())
      {
          var root = document.GetRootObject();
          // Style features, then assign to overlay
          overlay.document = document;
      }
  }
```

Under the hood, LoadFromUrlAsync wraps the callback-based LoadFromUrl in a TaskCompletionSource, so the await resumes on the main thread once the native async load completes. it works the same as coroutines but with a more concise syntax.

Source/Runtime/CesiumGeoJsonDocument.cs
baruchInsert-tech: Removed. The cleanup now happens in ~CesiumGeoJsonDocumentImpl on the native side, so the C# DisposeNative and finalizer are no longer needed.

native~/src/Runtime/CesiumGeoJsonDocumentRasterOverlayImpl.cpp
baruchInsert-tech: Removed

native~/src/Runtime/CesiumGeoJsonDocumentRasterOverlayImpl.cpp
baruchInsert-tech: Removed

Source/Runtime/CesiumGeoJsonObject.cs
baruchInsert-tech: The decision to use a single CesiumGeoJsonObject type was made to keep the Unity API simpler. Since the underlying native GeoJsonObject already exposes the type via GetObjectType(), users can check the type and access the relevant data (children, properties, geometry, etc.) without needing to cast between subtypes.

That said, I'm open to splitting it into separate subtypes to match the Unreal API if the team prefers consistency across platforms. Let me know if you'd like me to refactor it.

native~/src/Runtime/CesiumGeoJsonDocumentRasterOverlayImpl.cpp
baruchInsert-tech: Removed

Source/Runtime/CesiumGeoJsonObject.cs
j9liu: We do prefer consistency across platforms where possible, so if you're able to refactor this, that would be great!

baruchInsert-tech: Done

baruchInsert-tech:
@j9liu All requested changes were made and tested locally!
waiting for further instructions and/or wanted changes.

CHANGES.md
```diff
@@ -2,6 +2,18 @@
...
+
+- Added `CesiumGeoJsonDocumentRasterOverlay`, allowing stylized GeoJSON to be rasterized and draped over terrain and other 3D Tiles.
+- Added `CesiumGeoJsonDocument` to parse and manipulate GeoJSON documents in code.
+- Added `CesiumGeoJsonObject` to access individual features, geometries, and collections within a GeoJSON document.
+- Added `CesiumVectorStyle`, `CesiumVectorLineStyle`, `CesiumVectorPolygonStyle`, and `CesiumVectorPolygonFillStyle` structs for styling vector data.
+- Added support for per-feature styling, allowing different features in a single GeoJSON document to have different colors and styles.
+- Added `CesiumGeoJsonObjectType` enum for identifying GeoJSON object types (Point, LineString, Polygon, Feature, FeatureCollection, etc.).
+- Added `CesiumVectorColorMode` enum with `Normal` and `Random` color modes.
+- Added `CesiumVectorLineWidthMode` enum with `Pixels` and `Meters` width modes.
+- GeoJSON documents can be loaded from URLs, Cesium ion assets, or parsed directly from strings.
```
j9liu: This changelog can be slimmed down a bit; a lot of parts are interlinked and it's not worth mentioning every single one. But thanks for being thorough!

```suggestion
- Added `CesiumGeoJsonDocumentRasterOverlay`, allowing stylized GeoJSON to be rasterized and draped over terrain and other 3D Tiles.
- Added `CesiumGeoJsonDocument` to parse and manipulate GeoJSON documents in code.
- Added `CesiumGeoJsonObject` to access individual features, geometries, and collections within a GeoJSON document.
- Added `CesiumVectorStyle`, `CesiumVectorLineStyle`, `CesiumVectorPolygonStyle`, and `CesiumVectorPolygonFillStyle` structs for styling vector data.
```

Source/Runtime/ConfigureReinterop.cs
```diff
@@ -920,6 +920,104 @@ Cesium3DTilesetLoadFailureDetails tilesetDetails
...
+                go.GetComponent<CesiumGeoJsonDocumentRasterOverlay>();
+            geoJsonOverlay.source = geoJsonOverlay.source;
+            geoJsonOverlay.url = geoJsonOverlay.url;
+            geoJsonOverlay.ionAssetID = geoJsonOverlay.ionAssetID;
+            geoJsonOverlay.ionAccessToken = geoJsonOverlay.ionAccessToken;
+            geoJsonOverlay.ionServer = geoJsonOverlay.ionServer;
+            geoJsonOverlay.mipLevels = geoJsonOverlay.mipLevels;
+            CesiumVectorStyle vectorStyle = geoJsonOverlay.defaultStyle;
+            geoJsonOverlay.defaultStyle = vectorStyle;
+            CesiumGeoJsonDocument overlayDoc = geoJsonOverlay.GetDocumentInternal();
```
j9liu: This needs to be changed, or else it causes a compilation error.

```suggestion
            CesiumGeoJsonDocument overlayDoc = geoJsonOverlay.document;
```

native~/src/Runtime/CesiumGeoJsonObjectImpl.cpp
```diff
@@ -0,0 +1,492 @@
...
+
+  _pObject->getStyle() = std::nullopt;
+}
+
+void CesiumGeoJsonObjectImpl::DisposeNative(
+    const CesiumForUnity::CesiumGeoJsonObject& object) {
+  _pDocument = nullptr;
+  _pObject = nullptr;
+  _pStandaloneObject = nullptr;
+}
```
j9liu: Do this in destructor instead?

Source/Runtime/CesiumGeoJsonObject.cs
```diff
@@ -0,0 +1,144 @@
...
+        internal partial void DisposeNative();
+
+        /// <summary>
+        /// Finalizer to release native resources.
+        /// </summary>
+        ~CesiumGeoJsonObject()
+        {
+            DisposeNative();
+        }
+    }
```
j9liu: Is this necessary?

Source/Runtime/CesiumGeoJsonLineString.cs
```diff
@@ -0,0 +1,29 @@
...
+    /// </summary>
+    /// <remarks>
+    /// A LineString is a sequence of two or more positions forming a line.
+    /// Each point is a <see cref="double3"/> containing longitude, latitude,
+    /// and optional height (LLH) coordinates.
+    /// </remarks>
+    public class CesiumGeoJsonLineString
+    {
+        /// <summary>
+        /// Gets the points (LLH coordinates) that make up this line string.
```
j9liu: **NIT**: The `LLH` above `CesiumGeoJsonLineString` is fine because longitude/latitude/height is already spelled out, but the abbreviation here appears by itself. Better to spell it out.

```suggestion
        /// Gets the points in longitude, latitude, height that make up this line string.
```

Source/Runtime/CesiumGeoJsonFeatureIdType.cs
```diff
@@ -0,0 +1,23 @@
...
+
+        /// <summary>
+        /// The feature has a string ID.
+        /// </summary>
+        String = 1,
+
+        /// <summary>
+        /// The feature has an integer ID.
+        /// </summary>
+        Integer = 2
```
j9liu: This is flipped from Cesium for Unreal for some reason? Should be `None, Integer, String` in that order.

https://github.com/CesiumGS/cesium-unreal/blob/00cc5656cfa915f68fed5ad657559009c6339f69/Source/CesiumRuntime/Public/CesiumGeoJsonObject.h#L101

Source/Runtime/CesiumGeoJsonFeature.cs
```diff
@@ -0,0 +1,124 @@
...
+        /// </summary>
+        internal partial void DisposeNative();
+
+        /// <summary>
+        /// Finalizer to release native resources.
+        /// </summary>
+        ~CesiumGeoJsonFeature()
+        {
+            DisposeNative();
+        }
```
j9liu: Is this necessary?

native~/src/Runtime/CesiumGeoJsonFeatureImpl.cpp
```diff
@@ -0,0 +1,403 @@
...
+  }
+
+  _pFeature->style = std::nullopt;
+}
+
+void CesiumGeoJsonFeatureImpl::DisposeNative(
+    const CesiumForUnity::CesiumGeoJsonFeature& feature) {
+  _pDocument = nullptr;
+  _pFeature = nullptr;
+}
```
j9liu: Use destructor instead?

native~/src/Runtime/CesiumGeoJsonDocumentRasterOverlayImpl.cpp
```diff
@@ -0,0 +1,249 @@
...
+    : _pOverlay(nullptr) {}
+
+CesiumGeoJsonDocumentRasterOverlayImpl::
+    ~CesiumGeoJsonDocumentRasterOverlayImpl() {}
+
+void CesiumGeoJsonDocumentRasterOverlayImpl::AddToTileset(
+    const ::DotNet::CesiumForUnity::CesiumGeoJsonDocumentRasterOverlay& overlay,
+    const ::DotNet::CesiumForUnity::Cesium3DTileset& tileset) {
+  spdlog::default_logger()->info("CesiumGeoJsonDocumentRasterOverlay::AddToTileset called");
+
```
j9liu: Think this one got missed in the last review, whoops!

```suggestion
```

native~/src/Runtime/CesiumGeoJsonPolygonImpl.cpp
```diff
@@ -0,0 +1,62 @@
...
+#include <DotNet/Unity/Mathematics/double3.h>
+
+#include <cstdint>
+
+using namespace DotNet;
+using namespace CesiumVectorData;
+
+namespace CesiumForUnityNative {
+
+CesiumGeoJsonPolygonImpl::CesiumGeoJsonPolygonImpl(
```
j9liu: On the C# side, it's good that `CesiumGeoJsonPolygonImpl` is separate. On the C++ side, I'm not sure if this `Impl` has to exist. 

`pDocument` isn't doing anything here. I saw that in Unreal, `FCesiumGeoJsonPolygon` also references a document, but in hindsight, it's not doing anything too useful, and we should probably remove it there too.

Plus, the polygon rings can be stored as `double3`s on the Unity side, not necessarily `glm::dvec3`s. In this way, it's trivial to implement `GetPolygonRings` in C#. There is definitely work needed is to translate the rings from Native -> Unity, but given that a polygon can only come from `CesiumGeoJsonObject.GetObjectAsPolygon`, the work could be done there instead. Does that make sense?

Source/Runtime/ConfigureReinterop.cs
j9liu: This causes a compilation error.

native~/src/Runtime/CesiumGeoJsonObjectImpl.cpp
```diff
@@ -0,0 +1,492 @@
...
+    outline.width = style.polygonStyle.outlineStyle.width;
+    outline.widthMode =
+        fromUnityLineWidthMode(style.polygonStyle.outlineStyle.widthMode);
+    result.polygon.outline = outline;
+  }
+
+  return result;
+}
+
+} // namespace
```
j9liu: Some of the code here (mainly in converting between the Unity/Native enums, etc.) seems duplicate with `CesiumGeoJsonDocumentRasterOverlayImpl.cpp`. I would put what's needed in a separate file + helper function, then share it across the two.

native~/src/Runtime/CesiumGeoJsonFeatureImpl.cpp
```diff
@@ -0,0 +1,403 @@
...
+    outline.colorMode =
+        fromUnityColorMode(style.polygonStyle.outlineStyle.colorMode);
+    outline.width = style.polygonStyle.outlineStyle.width;
+    outline.widthMode =
+        fromUnityLineWidthMode(style.polygonStyle.outlineStyle.widthMode);
+    result.polygon.outline = outline;
+  }
+
+  return result;
+}
```
j9liu: See comment for duplicate style conversion code in `CesiumGeoJsonObjectImpl` and `CesiumGeoJsonDocumentRasterOverlayImpl`

Source/Runtime/CesiumVectorStyle.cs
```diff
@@ -0,0 +1,241 @@
...
+namespace CesiumForUnity
+{
+    /// <summary>
+    /// A simple 32-bit color struct that marshals correctly between C# and C++.
+    /// Unlike Unity's Color32, this struct does not use a union layout that
+    /// confuses Reinterop's code generation.
+    /// </summary>
+    [Serializable]
+    [StructLayout(LayoutKind.Sequential)]
+    public struct CesiumColor32
```
j9liu: I opened #665 so that `Color32` can be used without problems, so this `CesiumColor32` is no longer needed.

Source/Runtime/CesiumGeoJsonFeature.cs
```diff
@@ -0,0 +1,124 @@
...
+        /// </summary>
+        internal CesiumGeoJsonFeature()
+        {
+            CreateImplementation();
+        }
+
+        /// <summary>
+        /// Gets the type of ID on this feature.
+        /// </summary>
+        /// <returns>The ID type: None, String, or Integer.</returns>
```
j9liu: Once the enum order has been fixed...

```suggestion
        /// <returns>The ID type: None, Integer, or String.</returns>
```

Source/Runtime/CesiumGeoJsonFeature.cs
```diff
@@ -0,0 +1,124 @@
...
+        /// </summary>
+        /// <returns>The feature ID as a string, or an empty string if the
+        /// feature has no ID or the ID is not a string.</returns>
+        public partial string GetIdAsString();
+
+        /// <summary>
+        /// Gets the feature ID as an integer.
+        /// </summary>
+        /// <returns>The feature ID as a long, or 0 if the feature has no ID
+        /// or the ID is not an integer.</returns>
```
j9liu: Unreal uses `-1` instead of `0`. https://github.com/CesiumGS/cesium-unreal/blob/00cc5656cfa915f68fed5ad657559009c6339f69/Source/CesiumRuntime/Public/CesiumGeoJsonObject.h#L131-L139

Source/Runtime/CesiumGeoJsonFeature.cs
```diff
@@ -0,0 +1,124 @@
...
+        /// Gets the type of ID on this feature.
+        /// </summary>
+        /// <returns>The ID type: None, String, or Integer.</returns>
+        public partial CesiumGeoJsonFeatureIdType GetIdType();
+
+        /// <summary>
+        /// Gets the feature ID as a string.
+        /// </summary>
+        /// <returns>The feature ID as a string, or an empty string if the
+        /// feature has no ID or the ID is not a string.</returns>
```
j9liu: Unreal converts integer IDs to strings, even though they're not originally strings. And that seems to be what the `Impl` for this is doing anyway.

```suggestion
        /// <returns>The feature ID as a string, or an empty string if the
        /// feature has no ID. If the ID is an integer, it will be converted 
        /// to a string.</returns>
```

native~/src/Runtime/CesiumGeoJsonFeatureImpl.cpp
j9liu: Also, `_pDocument` isn't really doing anything here either. Remove?

native~/src/Runtime/CesiumGeoJsonObjectImpl.cpp
```diff
@@ -0,0 +1,492 @@
...
+      CesiumForUnity::CesiumVectorLineWidthMode::Pixels;
+
+  result.polygonStyle.fill = true;
+  result.polygonStyle.fillStyle.color.r = 255;
+  result.polygonStyle.fillStyle.color.g = 255;
+  result.polygonStyle.fillStyle.color.b = 255;
+  result.polygonStyle.fillStyle.color.a = 255;
+  result.polygonStyle.fillStyle.colorMode =
+      CesiumForUnity::CesiumVectorColorMode::Normal;
+  result.polygonStyle.outline = false;
```
j9liu: Can we just use `CesiumVectorStyle.Default` for this?

native~/src/Runtime/CesiumGeoJsonFeatureImpl.cpp
```diff
@@ -0,0 +1,403 @@
...
+    const CesiumForUnity::CesiumGeoJsonFeature& feature) {
+  if (!_pFeature || !_pFeature->geometry) {
+    return CesiumForUnity::CesiumGeoJsonObject(nullptr);
+  }
+
+  CesiumForUnity::CesiumGeoJsonObject result;
+  auto pGeomCopy =
+      std::make_shared<GeoJsonObject>(*_pFeature->geometry);
+  result.NativeImplementation().setNativeObject(std::move(pGeomCopy));
+  return result;
```
j9liu, line 219: What's the reason for making this copy instead of referencing it with a pointer? Since `CesiumGeoJsonFeatureImpl` has a `_pDocument` here, so we could pass that into `setNativeObjectInDocument` with the raw geometry pointer.

Source/Runtime/CesiumGeoJsonFeature.cs
```diff
@@ -0,0 +1,124 @@
...
+        /// <param name="style">The style to set.</param>
+        public partial void SetStyle(CesiumVectorStyle style);
+
+        /// <summary>
+        /// Clears any style set on this feature.
+        /// </summary>
+        /// <remarks>
+        /// After calling this method, the feature will use the default style.
+        /// </remarks>
+        public partial void ClearStyle();
```
j9liu: Is there an advantage to having these functions on both `CesiumGeoJsonFeature` as well as `CesiumGeoJsonObject`?

CHANGES.md
baruchInsert-tech: Done

Source/Runtime/ConfigureReinterop.cs
baruchInsert-tech: Done

native~/src/Runtime/CesiumGeoJsonObjectImpl.cpp
baruchInsert-tech: Done

Source/Runtime/CesiumGeoJsonObject.cs
baruchInsert-tech: Removed

Source/Runtime/CesiumGeoJsonLineString.cs
baruchInsert-tech: Done

Source/Runtime/CesiumGeoJsonFeatureIdType.cs
baruchInsert-tech: Done

Source/Runtime/CesiumGeoJsonFeature.cs
baruchInsert-tech: Removed

native~/src/Runtime/CesiumGeoJsonFeatureImpl.cpp
baruchInsert-tech: Removed _pDocument entirely from CesiumGeoJsonFeatureImpl

native~/src/Runtime/CesiumGeoJsonDocumentRasterOverlayImpl.cpp
baruchInsert-tech: Removed

native~/src/Runtime/CesiumGeoJsonPolygonImpl.cpp
baruchInsert-tech: Removed CesiumGeoJsonPolygonImpl entirely and converted CesiumGeoJsonPolygon to a pure C# class with a Rings property

native~/src/Runtime/CesiumGeoJsonObjectImpl.cpp
baruchInsert-tech: Done. Created CesiumVectorStyleConversions.h as a shared helper

native~/src/Runtime/CesiumGeoJsonFeatureImpl.cpp
baruchInsert-tech: Done

Source/Runtime/CesiumGeoJsonFeature.cs
baruchInsert-tech: Done

Source/Runtime/CesiumGeoJsonFeature.cs
baruchInsert-tech: Done, Updated both the doc comment and the C++ implementation to return -1 instead of 0 for the fallback, matching Cesium for Unreal.

Source/Runtime/CesiumGeoJsonFeature.cs
baruchInsert-tech: Done

native~/src/Runtime/CesiumGeoJsonObjectImpl.cpp
baruchInsert-tech: Done

native~/src/Runtime/CesiumGeoJsonFeatureImpl.cpp
baruchInsert-tech: Done, the geometry is now referenced by pointer instead of copied, with the document shared_ptr keeping it alive

Source/Runtime/CesiumGeoJsonFeature.cs
baruchInsert-tech: Removed style methods (HasStyle, GetStyle, SetStyle, ClearStyle) from CesiumGeoJsonFeature

Source/Runtime/CesiumVectorStyle.cs
baruchInsert-tech: I can wait for #665 to merge first, then replace CesiumColor32 with Color32 everywhere in my PR, or I can replace now assuming #665 will merge before mine.

j9liu: We're planning to merge #665 before this PR, so feel free to merge that into here + make your changes. Thanks!

Source/Runtime/CesiumGeoJsonPolygon.cs
```diff
@@ -8,45 +7,25 @@ namespace CesiumForUnity
...
         /// </remarks>
-        /// <returns>An array of <see cref="CesiumGeoJsonLineString"/> representing
-        /// the polygon's rings.</returns>
-        public partial CesiumGeoJsonLineString[] GetPolygonRings();
-
-        /// <summary>
-        /// Releases native resources associated with this polygon.
-        /// </summary>
-        internal partial void DisposeNative();
+        public CesiumGeoJsonLineString[] Rings { get; internal set; }
```
j9liu: We decapitalize properties (similar to Unity). Otherwise, the changes look great here, thank you!

```suggestion
        public CesiumGeoJsonLineString[] rings { get; internal set; }
```

Source/Runtime/CesiumGeoJsonLineString.cs
```diff
@@ -14,7 +14,7 @@ namespace CesiumForUnity
     public class CesiumGeoJsonLineString
     {
         /// <summary>
-        /// Gets the points (LLH coordinates) that make up this line string.
+        /// Gets the points in longitude, latitude, height that make up this line string.
         /// </summary>
         public double3[] Points { get; internal set; }
```
j9liu: Sorry I missed this earlier! We decapitalize properties (similar to Unity).

```suggestion
        public double3[] points { get; internal set; }
```

native~/src/Runtime/CesiumGeoJsonObjectImpl.h
j9liu: This can be removed now that we're no longer using it, right?

Source/Runtime/CesiumGeoJsonPolygon.cs
baruchInsert-tech: Done

Source/Runtime/CesiumGeoJsonLineString.cs
baruchInsert-tech: Done

native~/src/Runtime/CesiumGeoJsonObjectImpl.h
baruchInsert-tech: Done

j9liu (approved):
Thanks @baruchInsert-tech, this looks great! I'm just waiting for CI to pass before we merge, but I don't expect there to be any changes you have to do.

PR #664, opened 2026-02-25
Fix {y} and {reverseY} in UrlTemplateRasterOverlay documentation

timoore:
See https://github.com/CesiumGS/cesium-native/issues/1183


j9liu:
Thanks @timoore!

PR #676, opened 2026-04-04
Use new Unity licenses on CI

kring:
Every year we get new Unity licenses, which annoyingly have new entitlement IDs. So this PR updates CI to use them.


j9liu (approved):
Thank you @kring!!

PR #679, opened 2026-04-24
Fix broken CI caused by Unity moving the Unity Hub installer

kring:
Unity changed their URLs, which broke our CI. This fixes it.


j9liu:
Thanks @kring !

PR #689, opened 2026-05-07
Upgrade to vcpkg 2026.04.27 release

kring:
This depends on CesiumGS/cesium-native#1360 so merge that first.

This PR upgrades cesium-unity's vcpkg dependency stack to the [2026.04.27 release](https://github.com/microsoft/vcpkg/releases/tag/2026.04.27).

### cesium-native submodule
- Updated the cesium-native submodule to the `vcpkg-upgrade` branch, which bumps the vcpkg baseline and includes a Windows zlib fix.

### abseil overlay port → 20260107.1
- Customizations are identical to those in [cesium-unreal](https://github.com/CesiumGS/cesium-unreal/pull/1832): C++17 build, inline-namespace rename to avoid ODR violations with Unity's bundled abseil, suppressed `ABSL_LTS_RELEASE_VERSION` macros, forced `ABSL_OPTION_USE_STD_ORDERING=0` for Android NDK compatibility, and the cord.h NDK r25 fix.
- Two new upstream patch files (`fix-heterogeneous_lookup_testing-target.patch`, `fix-mingw-dll.patch`) are included.

### abseil UWP build fix
- Added a patch (`fix-uwp-time-zone-lookup.patch`) to fix an unresolved external symbol linker error on UWP targets. In abseil 20260107.1, `time_zone_lookup.cc` references `GetWindowsLocalTimeZone()` under `#if defined(_WIN32)`, which is true on UWP, but the function is only compiled when `CMAKE_SYSTEM_NAME == "Windows"`. The patch guards the call with `!WINAPI_FAMILY_APP` to match the compile-time guard.
- This is an upstream bug that we're working around here.

### sqlite3 overlay port → pinned to 3.51.3
- Added a sqlite3 overlay port pinned to version 3.51.3. SQLite 3.53.0 (the version in the new vcpkg baseline) intentionally dropped WinRT support.
- It's unclear if the sqlite3 developers intended to also drop UWP support (which we need for HoloLens, at least), but this is what happened.
- Version 3.51.3 is the latest release with UWP support via `SQLITE_OS_WINRT`.

### s2geometry overlay port (Android Clang 12 fix)
- Added an s2geometry overlay port with a patch that removes `[[nodiscard]]` from the `SpinLockHolder` class declaration. s2geometry 0.13.1 added `[[nodiscard]] ABSL_SCOPED_LOCKABLE` to this class, but mixing a C++ standard attribute with a GNU-style attribute before the class name triggers a Clang 12 bug (used by Unity's Android NDK r23) that treats the class as anonymous. Clang 14+ (used by Unreal's NDK r25) handles this correctly.

### blend2d overlay port removed
- Removed the blend2d overlay port as it is no longer needed.


kring:
Looks like the sqlite folks are aware of the regression and are moving toward fixing it: https://sqlite.org/forum/forumpost/b5a2f9413108b36b4e8b9e99d87c20118433e603f0ee3ececd77745c73374ef2

native~/vcpkg/ports/abseil/portfile.cmake
```diff
@@ -1,59 +1,172 @@
...
+# BACKGROUND — Unreal Engine and the Android NDK:
+#   Unreal Engine uses C++20. However, the Android NDK version bundled with
+#   Unreal has incomplete C++20 *library* support even though the compiler
+#   supports C++20 *language* features. This means headers that are compiled
+#   as part of Unreal's Android build (including installed vcpkg headers) must
+#   not rely on C++20 standard library types such as std::partial_ordering.
+#   We build abseil itself with C++17 to avoid any C++20 library dependencies
+#   in abseil's own compilation. We additionally patch the installed headers
+#   to be safe when included by Unreal's C++20 Android build.
+
```
j9liu, line 23: Change the text throughout this doc for Unity?

native~/vcpkg/ports/abseil/vcpkg.json
```diff
@@ -1,6 +1,6 @@
 {
   "name": "abseil",
-  "version": "20240722.0-cesium-for-unreal",
+  "version": "20260107.1-cesium-for-unreal",
```
j9liu, line 3: **NIT**: If it's not too much trouble...

```suggestion
  "version": "20260107.1-cesium-for-unity",
```

native~/vcpkg/ports/s2geometry/portfile.cmake
```diff
@@ -0,0 +1,63 @@
...
+#   ABSL_SCOPED_LOCKABLE expands to __attribute__((scoped_lockable)) on Clang.
+#   Clang 12 (used by Unity's Android NDK r23) has a bug where mixing a C++
+#   standard attribute ([[nodiscard]]) with a GNU-style attribute
+#   (__attribute__(...)) between the "class" keyword and the class name causes
+#   the class to be treated as anonymous, producing:
+#
+#     error: declaration of anonymous class must be a definition
+#     error: unknown type name 'SpinLockHolder'
+#
+#   Clang 14+ (Unreal Engine's Android NDK r25) handles this syntax correctly.
```
j9liu, line 26:
```suggestion
#   Clang 14+ (Unity's Android NDK r25) handles this syntax correctly.
```

j9liu (changes requested):
Thanks @kring for the PR! My only comments are about some copied references to "Unreal Engine" that should be changed to "Unity".

Otherwise, I spot checked the package in Unity and everything seems to work just as well.

native~/vcpkg/ports/abseil/portfile.cmake
kring: I intentionally made this an exact copy of the overlay port in Cesium for Unreal. If I had it to do over again, I would have put less Unreal-specific stuff in there in the first place. But I think the benefit of keeping the two in sync outweighs the weirdness of the comments. The commit message explains exactly where the overlay came from (a particular commit of cesium-unreal). This is useful because, when I told Copilot to do this update, I told it specifically to look at the commit messages to understand what changed and why so that you know how to incorporate the relevant changes into the new version. That worked really well.

native~/vcpkg/ports/s2geometry/portfile.cmake
kring: This time the reference to Unreal is actually correct! It's explaining why this overlay port is only needed in Unity, but not in Unreal.

j9liu:
Thanks @kring !

PR #691, opened 2026-05-13
Use 16KB alignment on Android

### LINKED ISSUE #618: Build native code with 16 KB page size on Android
Reported here:
https://community.cesium.com/t/16-kb-memory-page-size-for-unity3d-android-project/43597

The Google Play Store will soon require that all published apps are built using a 16 KB page size. We should modify our build process to do so.

kring:
Fixes #618

### DIFF
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -1,5 +1,11 @@
 # Change Log {#changes}
 
+## ? - ?
+
+##### Fixes :wrench:
+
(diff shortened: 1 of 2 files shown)


kring:
Based on the advice here: https://developer.android.com/guide/practices/page-sizes#cmake

j9liu (approved):
Thanks @kring! I wish I had the environment to test this, but the change looks straightforward enough to merge.

PR #695, opened 2026-05-22
Add Force Double Sided toggle to Cesium3DTileset

triktron:
## Description

Adds an opt-in `Force Double Sided` boolean to `Cesium3DTileset`. When enabled, every per-primitive material has `_DoubleSidedEnable = 1` and `_Cull / _CullMode / _BUILTIN_CullMode = Off`, applied **after** the per-primitive glTF material values are written.

This makes it possible to assign a custom `Opaque Material` with `Render Face = Both` (URP) or its built-in / HDRP equivalent and have that setting actually stick. Currently, the cull state on every cloned material is overwritten on each tile load from `gltfMaterial.doubleSided` (see #491 for why that override exists), so user-supplied materials with double-sided rendering silently revert to back-face culling.

Default is `false`, so the existing glTF-spec-driven behavior introduced in #491 is unchanged for anyone who doesn't opt in.

**Where the override is applied (native):** outside the existing `if (pMaterial) { setGltfMaterialParameterValues(...); }` block in `UnityPrepareRendererResources.cpp`, so it runs after the glTF write and also covers primitives that have no glTF material attached.

**How the C# property reaches native code:** standard Reinterop wiring — `_forceDoubleSided` field + `forceDoubleSided` property on `Cesium3DTileset`, an exposure line in `ConfigureReinterop.cs`, and `tilesetComponent.forceDoubleSided()` at the native call site. No header changes — Reinterop generates the binding.

## Issue number or link

No tracking issue — small additive feature. Happy to open one if preferred.

## Author checklist

## Testing plan

Manual, in both URP and Built-in (have not yet verified HDRP, where the original #491 had quirks).

1. Fresh scene, `Cesium3DTileset` with `Opaque Material = None`, `Force Double Sided = false` — default behavior preserved:
   - Tiles whose glTF has `doubleSided = false` are back-face culled.
   - Tiles whose glTF has `doubleSided = true` render from both sides.
2. Same scene, flip `Force Double Sided = true` — every tile renders both sides. Runtime per-primitive material has `_Cull == 0`, `_CullMode == 0`, `_DoubleSidedEnable == 1`.
3. Duplicate `CesiumDefaultTilesetMaterial`, set `Render Face = Back` so `_Cull = 2`, assign to `Opaque Material`, toggle `Force Double Sided = true` — Cesium still renders double-sided (verifies the toggle wins over the source material as well).
4. Same custom material, `Force Double Sided = false` — cull state matches the glTF (original behavior).
5. Inspector: the new "Force Double Sided" entry appears in the Render section between "Generate Smooth Normals" and "Ignore KHR_materials_unlit", with the tooltip on hover.

### DIFF
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -2,6 +2,10 @@
 
 ## ? - ?
 
+##### Additions :tada:
+
+- Added a `Force Double Sided` toggle to `Cesium3DTileset`. When enabled, all tile primitives are rendered with back-face culling disabled, regardless of the glTF material's `doubleSided` flag or the cull state of a user-supplied `Opaque Material`.
+
 ##### Fixes :wrench:
 
 - The native Android library (`libCesiumForUnityNative.so`) is now built with 16 KB ELF segment alignment, enabling compatibility with Android 15+ devices that use 16 KB memory pages.
--- a/Source/Editor/Cesium3DTilesetEditor.cs
+++ b/Source/Editor/Cesium3DTilesetEditor.cs
@@ -36,6 +36,7 @@ namespace CesiumForUnity
         //private SerializedProperty _useLodTransitions;
         //private SerializedProperty _lodTransitionLength;
         private SerializedProperty _generateSmoothNormals;
+        private SerializedProperty _forceDoubleSided;
 
         private SerializedProperty _pointCloudShading;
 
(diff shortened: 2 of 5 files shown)


triktron:
thinking about the problem, perhaps its better if we don't force double-sided on, but instead add a toggle that forced the material properties over the gltf. I'll modify the code later this week.

PR #699, opened 2026-05-28
Fix reference to GeoJsonDocumentRasterOverlay

azrogers:
Quick fix to let Cesium for Unity build with the vector tiles branch.

This should only be merged in after CesiumGS/cesium-native#1365 is.


j9liu (approved):

j9liu:
Thanks @azrogers!

PR #705, opened 2026-07-01
Fixed TokenTroubleshooting for CesiumGeoJsonDocumentRasterOverlay

gojushin:
## Description

tldr: 

Clicking "Troubleshoot Token" on a `CesiumGeoJsonDocumentRasterOverlay` Component causes a `NullReferenceException`, because of a missing Implementation.

non-tldr:

The main root of the issue lies in a missing implementation inside the `getServer` function of `IonTokenTroubleshootingWindowImpl.cpp`.

When the `CesiumIonAsset` used in the window is created, this constructor is called:
```csharp
public CesiumIonAsset(CesiumGeoJsonDocumentRasterOverlay overlay)
{
    this._type = AssetType.GeoJsonOverlay;
    this._geoJsonOverlay = overlay;
}
```
Consequently, `getServer` returns a `nullptr` in the final else block because both `asset.tileset` and `asset.overlay` are never set.
However, the constructor in `CesiumIonAsset.cs` specifically sets the `_geoJsonOverlay` variable (which has a public getter), which we can use here.

(This follows the same pattern as `_tileset` and `_overlay`)

So. Eventually, the issue can be fixed by adding a check for `asset.geoJsonOverlay` (and adding it to `ConfigureReinteropEditor.cs`'s `ExposeToCPP` of course).

## Issue number or link

I very briefly mentioned this bug [here](https://community.cesium.com/t/introducing-efficient-geojson-cutouts/46510/11)

## Author checklist

_(However, I wasn't sure where to put it, so I just added it to the latest changes....😅)_

## Testing plan

In the current version of the plugin clicking "Troubleshoot Token" on a `CesiumGeoJsonDocumentRasterOverlay` Component causes a `NullReferenceException`.

After implementing the fix, everything works as expected.

### DIFF
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -14,6 +14,9 @@
 
 In addition to the above, this release updates [cesium-native](https://github.com/CesiumGS/cesium-native) from v0.61.0 to v0.62.0. See the [changelog](https://github.com/CesiumGS/cesium-native/blob/main/CHANGES.md) for a complete list of changes in cesium-native.
 
+##### Fixes :wrench:
+- Fixed a `NullReferenceException` thrown when opening the Cesium ion token troubleshooting window for a `CesiumGeoJsonDocumentRasterOverlay`.
+
 ## v1.23.3 - 2026-06-01
 
 ##### Fixes :wrench:
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -6,6 +6,10 @@
 
 - Added support for the Linux platform (x86-64 only).
 
+##### Fixes :wrench:
+
+- Fixed a `NullReferenceException` thrown when opening the Cesium ion token troubleshooting window for a `CesiumGeoJsonDocumentRasterOverlay`.
+
 ## v1.24.0 - 2026-07-01
 
 ##### Additions :tada:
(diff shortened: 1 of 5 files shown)


CHANGES.md
```diff
@@ -8,6 +8,9 @@
 
 In addition to the above, this release updates [cesium-native](https://github.com/CesiumGS/cesium-native) from v0.61.0 to v0.62.0. See the [changelog](https://github.com/CesiumGS/cesium-native/blob/main/CHANGES.md) for a complete list of changes in cesium-native.
 
+##### Fixes :wrench:
```
j9liu: Can you merge in main and move the changelog entry under a new section?

Also a nitpick, but leave a blank line between "Fixes" and the entry.

j9liu (commented):
Thank you for the fix @gojushin ! I confirmed that the bug occurs on `main` but does not occur on this branch.

Just one note about updating the changelog -- sorry it took a while to get to this!

CHANGES.md
gojushin: Done ^^

j9liu (approved):
Thanks @gojushin ! Just waiting on CI to pass before we merge.

PR #706, opened 2026-07-09
Add GeoJson as source for CesiumCartographicPolygon

gojushin:
## Description

This PR introduces efficient cutouts using  `.geoJson`  files.

It builds on the functionality introduced in [this](https://github.com/CesiumGS/cesium-unity/pull/656) PR and primarily extends  `CesiumCartographicPolygon`  to support loading a  `Spline`  Component from a  `.geoJson`  file.

Using `CesiumCartographicPolygon`  instead of `CesiumGeoJsonDocumentRasterOverlay` for hiding the tiles has a significant advantage: tiles that are fully contained within the polygon are not just visually hidden, but are never loaded in the first place, resulting in a net performance improvement in these scenarios.

While implementing this feature, I realized it would be beneficial for the workflow to support both directions ( `.geoJson` ↔  `Spline`)... So I also added the ability to export Splines as  `.geoJson`  files, allowing them to be further integrated into the Cesium ecosystem. _This can however be easily taken out again, if the functionality is undesired._

image

## Issue number or link

This implements the functionality discussed here:
https://community.cesium.com/t/introducing-efficient-geojson-cutouts/46510

## Author checklist

## Remaining Tasks

None

## Testing plan

I tested my changes using .geoJson files created in QGIS for each of the 3 new modes (FromDocument, FromURL and FromIon).
The .geoJson Writing has been tested by opening it in QGIS.


j9liu:
Thanks for the PR @gojushin! Confirmed that we have a CLA from you on hand, we'll take a look at this when we get the chance. 🙏

j9liu:
Sorry for the delay on this @gojushin! Thanks for following our PR checklist thoroughly by updating the changelog/unit tests 🙌 

I have a few things to discuss before we merge -- let me know your thoughts!

### Hardcoded polygon access in GeoJSON

Currently there's assumptions made about the structure of the GeoJSON in order to extract a usable polygon ring. It basically defaults to the first one it can find -- which is not a bad solution, but I think it offers less flexibility as a result of being hardcoded.

```c++
            else if (objType == CesiumGeoJsonObjectType.FeatureCollection)
            {
                CesiumGeoJsonFeature[] features = obj.GetObjectAsFeatureCollection();
                if (features == null || features.Length == 0)
                    return null;
                geometry = features[0].GetGeometry();
            }


            if (geometry == null)
                return null;


            CesiumGeoJsonLineString[] rings = null;


            CesiumGeoJsonObjectType geometryType = geometry.GetObjectType();
            if (geometryType == CesiumGeoJsonObjectType.Polygon)
            {
                rings = geometry.GetObjectAsPolygon()?.rings;
            }
            else if (geometryType == CesiumGeoJsonObjectType.MultiPolygon)
            {
                CesiumGeoJsonPolygon[] polys = geometry.GetObjectAsMultiPolygon();
                rings = polys?[0]?.rings;
            }
```

I don't want this to be a deal breaker, just wondering if there's any way to improve this. Otherwise we'll have to be sure to note this assumption in the documentation.

### Saving polygon assets

There are implications with streaming Cesium ion assets -- given that access can be dynamic due to creating/revoking new access tokens, I don't know if it's right for us to allow users to save assets that they stream in. GeoJSONs can be potentially loaded as Cesium ion assets using an ion asset ID and token.

With that in mind, let's remove the save-to-disk functionality for now. If it ends up being okay after all, then I'll leave a comment for it and we can do a follow-up PR. But I think it will be unlikely. Sorry about that!

gojushin:
@j9liu 

Regarding the assumption of the first ring: 

That's defined as per [RFC7946 spec](https://datatracker.ietf.org/doc/html/rfc7946):

>  For Polygons with more than one of these rings, the first MUST be
>       the exterior ring, and any others MUST be interior rings.  The
>       exterior ring bounds the surface, and the interior rings (if
>       present) bound holes within the surface.

And the exterior ring is _usually_ what we assume for cutouts.
I believe that is also how it is currently done in CesiumJS when we load a geojson as ClippingPolygon (please correct me if I am wrong!!) - although there the holes DO remain accessible as an array of arrays.

It would ofc be easy to expose an index. 

OR do it a simlarish way CesiumJS does! (exposing holes as extra field https://cesium.com/learn/ion-sdk/ref-doc/ClippingPolygon.html has that as option in the constructor)
`CesiumCartographicPolygon` feels like the right place to add this.

Going into the future I would in any case attempt to go down the route of introducing [ClippingPolygonCollection](https://cesium.com/learn/ion-sdk/ref-doc/ClippingPolygonCollection.html)'s. Which this should be a neat foundation for ^^.

As for saving geojsons to disk: 

We can remove that ^^ Was only a suggestion :P.

Let me know what you can think! I can do the required refactoring around mid next week.

gojushin:
@j9liu I have removed the saving functionality and added a dynamic index for both FeatureCollections & MultiPolygons.

Sorry for talking a bit of nonsense in my earlier reply, I needed to pick the code back up to actually understand what you meant xD. 

Anyhow. The index is no longer hardcoded and tooltips have been added to guide users on how the functionality works.

tldr:
> - Select FeatureCollection via Index
> - Select Polygon of the selected FeatureCollection via Index.
> - If any index is out of range, the Cutout gets cleared instead.
> - Changing the Index in Editor, rebuilds the Spline automatically.

It would probably make sense to add this to the documentation.... I do however not know how and where to do that :P
If you let me know how, I will gladly update it.

cescutout

I hope that's what you had in mind ^^

Also regarding my earlier proposal to go down the route of [ClippingPolygonCollection](https://cesium.com/learn/ion-sdk/ref-doc/ClippingPolygonCollection.html).
I also noticed that this is not necessary, as `CesiumPolygonRasterOverlay` already takes multiple `CesiumCartograhpicPolygon`'s. So it's pretty much all there now.

CHANGES.md
```diff
@@ -5,6 +5,7 @@
 ##### Additions :tada:
 
 - Added `CesiumVectorPointStyle` and a `pointStyle` field on `CesiumVectorStyle`, enabling `CesiumGeoJsonDocumentRasterOverlay` to render GeoJSON `Point` and `MultiPoint` features.
+- Added support for loading `.geoJson` files to `CesiumCartographicPolygon`.
```
j9liu, line 8: **NIT**:

```suggestion
- Added support for loading `CesiumCartographicPolygon`s from polygons in GeoJSON files.
```

Source/Runtime/CesiumCartographicPolygon.cs
```diff
@@ -8,10 +9,45 @@
...
+        /// <summary>
+        /// The polygon's shape is loaded from a GeoJSON document hosted on Cesium ion.
+        /// </summary>
+        FromDocument = 1,
+
+        /// <summary>
+        /// The polygon's shape is loaded from a GeoJSON document that has been parsed
+        /// and assigned in code using <see cref="CesiumCartographicPolygon.document"/>.
+        /// </summary>
+        FromCesiumIon = 2,
```
j9liu, line 33: These two comments are flipped.

Source/Runtime/CesiumCartographicPolygon.cs
```diff
@@ -8,10 +9,45 @@
...
+        /// <summary>
+        /// The polygon's shape is defined manually by editing the spline directly
+        /// in the Editor (or at runtime).
+        /// </summary>
+        Manual = 0,
+
+        /// <summary>
+        /// The polygon's shape is loaded from a GeoJSON document hosted on Cesium ion.
+        /// </summary>
+        FromDocument = 1,
```
j9liu, line 27: Adding `GeoJson` helps with clarity in the dropdown here, since we can't see any tooltips that explain the enum:

Image

```suggestion
        FromGeoJsonDocument = 1,
```

Source/Runtime/CesiumCartographicPolygon.cs
```diff
@@ -84,7 +343,244 @@ void Reset()
...
+                    if (this._ionAssetID > 0)
+                        loaded = await CesiumGeoJsonDocument.LoadFromCesiumIonAsync(
+                            this._ionAssetID, this._ionAccessToken, this.ionServer);
+                    break;
+            }
+
+            if (loaded == null && this._source != CesiumCartographicPolygonSource.Manual)
+            {
+                Debug.LogWarning("CesiumCartographicPolygon: failed to load the GeoJSON. The spline was left unchanged.");
+            }
```
j9liu, line 412: Usually when Cesium ion assets fail to load, we open a ion troubleshooting window for the asset. Could we do that here?

Source/Runtime/CesiumCartographicPolygon.cs
```diff
@@ -84,7 +343,244 @@ void Reset()
...
+            if (rings == null || rings.Length == 0 || rings[0] == null || rings[0].points.Length < 3)
+                return null;
+
+            List<double2> result = new List<double2>(rings[0].points.Length);
+            foreach (double3 p in rings[0].points)
+                result.Add(new double2(p.x, p.y));
+            return result;
+        }
+
+        private void ApplyDocument(CesiumGeoJsonDocument geoJsonDocument)
```
j9liu, line 505: Whiel testing, I found that if I used a bogus ion Asset ID to load this (for example, a non-GeoJSON asset), then the original spline will still stay around. It should clear the spline points to reflect that the new asset isn't working. I'm not sure where this needs to happen in the code, but if you can ensure that the spline clears in this case, that would be great :pray:

j9liu (commented):
This is working great @gojushin! Thanks for all your work on this and the requested changes. :pray:

I have a few small comments that are mostly small functionality or documentation tweaks. Once those are addressed, this should be good to merge!

Was able to test this using the ion assets in this CesiumJS [Sandcastle](https://sandcastle.cesium.com/?id=aec-clipping) -- nicely done!

Image

CHANGES.md
gojushin: Adjusted!

Source/Runtime/CesiumCartographicPolygon.cs
gojushin: Good catch! Fixed

Source/Runtime/CesiumCartographicPolygon.cs
gojushin: Done!

Source/Runtime/CesiumCartographicPolygon.cs
gojushin: Absolutely.

I implemented it in the same way that `RasterOverlay`'s and `3DTileset`'s handle it.

That being said, `CesiumGeoJsonDocument.LoadFromCesiumIonAsync` doesn't expose HTTP status codes, so the handler can't filter on HTTP status codes, like the others do.

Therefore `IonTokenTroubleshootingWindow` will open regardless of that.

Implementing filtering by HTTP status code would need us to not discard it in `CesiumGeoJsonDocumentImpl.cpp` L111-L126 (and then pass it along to C#). _(Read more below)_

I documented this in code.

Source/Runtime/CesiumCartographicPolygon.cs
gojushin: This should already be happening, when manually clicking refresh. So far the code just didn't auto-refresh when a new id/url was entered.

It does now.

gojushin:
As a result of implementing `IonTokenTroubleshootingWindow` auto-opening without status-code filtering and also implementing the auto-refreshing, using wrong datasets (e.g. an asset ID pointing to a 3DTileset instead of a GeoJSON document) causes the troubleshooting window to open and report "everything OK" across the board while nothing indicated that the wrong type of asset was being loaded.

In short: To implement this cleanly we NEED status codes.

`GeoJsonDocument::fromCesiumIonAsset` and `GeoJsonDocument::fromUrl` do however never surface them, besides baking them into a ErrorMessage in `ErrorList` (see: `GeoJsonDocument.cpp` L639-L641). It would be cleanest to expose them in Native first.....

But I'll be honest. I am not familiar enough with Native and all the implementations that depend on it to implement that on my own confidently.

For testing I replaced `GeoJsonDocument::fromUrl`/`fromCesiumIonAsset` in `CesiumGeoJsonDocumentImpl.cpp` with a hand-rolled version that follows the same pattern from the Native implementation and added the appropriate carry through to C# and it works flawlessly.

-----

@j9liu I am a little lost on where to continue now.
I can commit the version that works, but it violates DRY, I could wait for a fix in Native , or we could roll as-is and fix this small thing in a coming PR. Your call ^^!

-----

A small Sidenote: I noticed that the documentation for `IAssetResponse` states:
 
> A completed response for a 3D Tiles asset.

which is technically false. It carries data also for Ion Assets in general and is used throughout the codebase like that. Perhaps that could need a tiny adjustment :D

j9liu:
Ah, I didn't consider how the troubleshooting window would be difficult to implement. Sorry @gojushin! :(

Yes, let's roll with this as-is and fix it in a follow-up PR. I could take a look into how we surface the error codes when I get the chance.

Let me know if I should do another review. Thanks for addressing the feedback so thoroughly!

gojushin:
No worries at all @j9liu! The codebase is an absolute joy to work with. (●'◡'●)

Sounds good to me! Feel free to open an issue and assign me to it so that we don't forget to add this once Native has what we need!

I don't think another review is necessary, as the implementation closely mirrors [the existing one](https://github.com/CesiumGS/cesium-unity/blob/main/Source/Runtime/Cesium3DTileset.cs#L63-L83). But it can ofc never hurt if you'd have another swift look!

Thanks for your work and input!

Issue #707, opened 2026-07-13
CesiumGeoJsonDocumentRasterOverlay does not render GeoJSON Point / MultiPoint features


baruchInsert-tech:
### Description

`CesiumGeoJsonDocumentRasterOverlay` renders `LineString` and `Polygon` geometry correctly, but GeoJSON `Point` and `MultiPoint` features are silently dropped - nothing is drawn for them.

There is also no managed way to configure point appearance: `CesiumVectorStyle` exposes `lineStyle` and `polygonStyle`, but there is no point style.

### Steps to reproduce

1. Add a `Cesium3DTileset` to a scene (e.g. Cesium World Terrain).
2. Add a `CesiumGeoJsonDocumentRasterOverlay` component to it and point it at a GeoJSON document that contains `Point` / `MultiPoint` features (via URL, or a `CesiumGeoJsonDocument`).
3. Enter Play mode and view the tileset.

A minimal document that reproduces it:

```json
{
  "type": "FeatureCollection",
  "features": [
    { "type": "Feature", "properties": {},
      "geometry": { "type": "Point", "coordinates": [-75.61, 40.04] } },
    { "type": "Feature", "properties": {},
      "geometry": { "type": "LineString", "coordinates": [[-75.62, 40.04], [-75.60, 40.05]] } }
  ]
}
```

### Expected behavior

`Point` and `MultiPoint` features are rasterized onto the tileset (as filled and/or outlined circles), consistent with how cesium-native's `VectorRasterizer` already draws points, and configurable through a point style on `CesiumVectorStyle`.

### Actual behavior

`Point` / `MultiPoint` features are not drawn. `LineString` / `Polygon` features in the same document render as expected.

### Analysis / root cause

The rasterization side is already handled in cesium-native - the gap is entirely on the cesium-unity side.

1. **cesium-native (rasterization) - already done.** `VectorRasterizer::drawPoints` and `PointStyle` already exist, and `GeoJsonDocumentRasterOverlay` on cesium-native `main` now rasterizes `Point` / `MultiPoint` geometry (it derives a point footprint from `PointStyle::radius` so points are no longer filtered out before reaching the rasterizer). No further native work is required.

[...]

PR #708, opened 2026-07-13
Add CesiumVectorPointStyle for GeoJSON point rendering

baruchInsert-tech:
@j9liu 
@azrogers 

image

## Description

Exposes a managed point style so `CesiumGeoJsonDocumentRasterOverlay` can render GeoJSON `Point` and `MultiPoint` features, which were previously dropped.

## Changes

- Add `CesiumVectorPointStyle` (radius, fill + fill style, outline + outline style) and a `pointStyle` field on `CesiumVectorStyle`. The default enables fill, since the native `PointStyle` default (no fill, no outline) renders nothing.
- Wire `pointStyle` to/from the native `VectorStyle::point` in `CesiumVectorStyleConversions.h`, mirroring the existing polygon style.
- Exercise the new type in `ConfigureReinterop.cs` so Reinterop generates its interop header.
- Add point-style tests to `TestCesiumVectorStyle.cs`.
- Changelog entry.

The overlay's inspector needs no change - `CesiumGeoJsonDocumentRasterOverlayEditor` draws the style with `PropertyField(..., includeChildren: true)`, so the new **Point Style** section appears automatically.

## Native dependency

This PR is the managed side only and **intentionally does not bump the cesium-native submodule**.

It compiles against the currently pinned native (`PointStyle` and `VectorStyle::point` already exist there), but **points will not actually render until the cesium-native submodule is advanced to a version that includes the `GeoJsonDocumentRasterOverlay` point-rasterization support** - which is already merged in cesium-native `main`. Until that bump lands, `pointStyle` is configurable but has no visible effect.

Suggested follow-up: bump the submodule to a cesium-native version containing that support (coordinated separately / with the next native update).

## Issue number or link

https://github.com/CesiumGS/cesium-unity/issues/707

## Author checklist

## Remaining Tasks

Bump the submodule to a cesium-native version containing the `GeoJsonDocumentRasterOverlay` point-rasterization support.

## Testing plan

- Added unit tests for `CesiumVectorPointStyle` defaults and configuration.
- Verified end-to-end locally against a cesium-native build that includes the point-rasterization support: `Point`/`MultiPoint` features from a GeoJSON document rendered onto a tileset via `CesiumGeoJsonDocumentRasterOverlay`.

## Reviewer checklist

Thank you for taking the time to review this PR. By approving a PR you are taking as much responsibility for these changes as the author.

As you review, please go through the checklist below:

- [ ] Review and run all parts of the test plan on this branch and verify it matches expectations.
    - [ ] If the issue is a bug please make sure you can reproduce the bug in the main branch and then checkout this branch to make sure it actually solved the issue.
- [ ] Review the code and make sure you do not have any remaining questions or concerns. You should understand the code change and the chosen approach. If you are not confident or have doubts about the code, please do not hesitate to ask questions.
    - [ ] Code should follow the [C++ Style Guide](https://cesium.com/learn/cesium-native/ref-doc/style-guide.html)
- [ ] Review the unit tests and make sure there are no missing tests or edge cases.
- [ ] Review documentation changes and updates to `CHANGES.md` to make sure they accurately cover the work in this PR.
- [ ] Verify that the [Contributor License Agreement](https://github.com/CesiumGS/community/tree/main/CLAs) has been submitted, if needed.

### DIFF
--- a/Tests/TestCesiumVectorStyle.cs
+++ b/Tests/TestCesiumVectorStyle.cs
@@ -106,46 +106,6 @@ public class TestCesiumVectorStyle
 
     #endregion
 
-    #region CesiumVectorPointStyle Tests
-
-    [Test]
-    public void PointStyleDefaultHasFillEnabledOutlineDisabled()
-    {
-        CesiumVectorPointStyle style = CesiumVectorPointStyle.Default;
-
-        Assert.AreEqual(5.0, style.radius, 0.001);
-        Assert.IsTrue(style.fill);
-        Assert.IsFalse(style.outline);
-    }
-
-    [Test]
-    public void PointStyleCanEnableBothFillAndOutline()
-    {
-        CesiumVectorPointStyle style = new CesiumVectorPointStyle();
-        style.radius = 8.0;
-        style.fill = true;
-        style.fillStyle = CesiumVectorPolygonFillStyle.Default;
-        style.outline = true;
-        style.outlineStyle = CesiumVectorLineStyle.Default;
-
-        Assert.AreEqual(8.0, style.radius, 0.001);
-        Assert.IsTrue(style.fill);
-        Assert.IsTrue(style.outline);
-    }
-
-    [Test]
-    public void PointStyleCanDisableBothFillAndOutline()
-    {
-        CesiumVectorPointStyle style = new CesiumVectorPointStyle();
-        style.fill = false;
-        style.outline = false;
-
-        Assert.IsFalse(style.fill);
-        Assert.IsFalse(style.outline);
-    }
-
-    #endregion
-
     #region CesiumVectorStyle Tests
 
     [Test]
--- a/Tests/TestCesiumVectorStyle.cs
+++ b/Tests/TestCesiumVectorStyle.cs
@@ -106,6 +106,46 @@ public class TestCesiumVectorStyle
 
     #endregion
 
+    #region CesiumVectorPointStyle Tests
+
+    [Test]
+    public void PointStyleDefaultHasFillEnabledOutlineDisabled()
+    {
+        CesiumVectorPointStyle style = CesiumVectorPointStyle.Default;
+
+        Assert.AreEqual(5.0, style.radius, 0.001);
+        Assert.IsTrue(style.fill);
+        Assert.IsFalse(style.outline);
+    }
+
+    [Test]
+    public void PointStyleCanEnableBothFillAndOutline()
+    {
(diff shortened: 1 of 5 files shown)


baruchInsert-tech:
### Note on the point fill/outline style types

`CesiumVectorPointStyle.fillStyle` reuses the existing `CesiumVectorPolygonFillStyle` (and `outlineStyle` reuses `CesiumVectorLineStyle`) rather than introducing point-specific types. This is deliberate:

- **It matches the native model.** In cesium-native, a point's fill is a `ColorStyle` and
  its outline is a `LineStyle` - the same building blocks polygons use. There's no
  point-specific fill/outline concept to mirror, so a distinct managed type would just
  duplicate `CesiumVectorPolygonFillStyle` field-for-field.
- **It matches cesium-unreal.** `FCesiumVectorPointStyle` reuses the polygon fill style the
  same way, so the two engines stay consistent.
- **It keeps this PR scoped.** Renaming `CesiumVectorPolygonFillStyle` to a neutral
  `CesiumVectorFillStyle` would touch polygon styling as well, change the interop surface
  (requiring Reinterop regeneration + a native rebuild), and turn an additive "render points"
  change into a shared-type rename.

I'm happy to introduce a generic `CesiumVectorFillStyle` (aliasing/replacing the polygon one)
as a follow-up if you'd prefer the naming to be geometry-neutral - I left it out here to keep
the diff minimal and additive. Let me know your preference.

baruchInsert-tech:
@j9liu 
Hi,
any update on this PR?

Tests/TestCesiumVectorStyle.cs
```diff
@@ -106,6 +106,46 @@ public void PolygonStyleCanDisableBothFillAndOutline()
...
+    {
+        CesiumVectorPointStyle style = new CesiumVectorPointStyle();
+        style.fill = false;
+        style.outline = false;
+
+        Assert.IsFalse(style.fill);
+        Assert.IsFalse(style.outline);
+    }
+
+    #endregion
```
azrogers, line 107: These tests are kinda pointless. They're basically just testing if variable assignment does in fact work. I don't think we need them.

azrogers (changes requested):

Tests/TestCesiumVectorStyle.cs
baruchInsert-tech: removed

azrogers:
Thanks @baruchInsert-tech!

PR #717, opened 2026-07-31
Move Android build from Windows job to Linux job in CI

kring:
## What
Moves the Android player build from the Windows CI job to the Linux CI job.

## Why
Windows was consistently the longest-running job in CI, in large part because it also built the Android target (SDK/NDK module install + native compile) on top of Windows/UWP. Android's native build logic is already OS-agnostic (`CompileCesiumForUnityNative.cs`/`Build~/Package.cs` just set up the Android NDK toolchain regardless of host), and Android builds directly on the host rather than in the Linux job's `almalinux:8` container, so moving it to Linux lets Windows and Linux build parallel workloads more evenly instead of stacking three platforms onto Windows.

## Changes
- **Windows job**: drop the `android`/`android-sdk-ndk-tools`/`android-open-jdk-11.0.14.1+1` Unity modules, drop `--platform Android`, remove the now-unneeded `ANDROID_NDK_ROOT` clearing step.
- **Linux job**: install `nasm` (needed for vcpkg asm-optimized ports on the host, previously only installed inside the container), install the Android Unity modules (with `--childModules` so `unityhub` doesn't block on an interactive prompt to confirm installing the NDK submodule), unset `ANDROID_NDK_ROOT` so Unity's bundled NDK is used, and add `--platform Android` to the build command.
- No source (`.cs`) changes were needed - the native build logic was already OS-agnostic for Android.

## Verification
- Pushed the branch and ran the full CI workflow to completion: https://github.com/CesiumGS/cesium-unity/actions/runs/30605283921 (all jobs green).
- Downloaded and inspected the actual artifact contents (not just job status):
  - Linux Package now contains `Plugins/Android/{arm64,x86_64}/libCesiumForUnityNative.so`.
  - Combined Package still contains both Android `.so` files, correctly merged in.
  - Windows Package shrank (~38MB vs. previously much larger) and contains no Android files.
- PlayMode "Run Tests"/"Test Report" steps passed in both the Linux and Windows jobs.

*Opened by GitHub Copilot on behalf of @kring.*

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -77,7 +77,7 @@ jobs:
       - name: Install Unity
         run: |
           Start-Process -FilePath "C:/Program Files/Unity Hub/Unity Hub.exe" -Args "-- --headless install --version 2022.3.41f1 --changeset 0f988161febf" -Wait
-          Start-Process -FilePath "C:/Program Files/Unity Hub/Unity Hub.exe" -Args "-- --headless install-modules --version 2022.3.41f1 --changeset 0f988161febf --module android android-sdk-ndk-tools android-open-jdk-11.0.14.1+1 universal-windows-platform" -Wait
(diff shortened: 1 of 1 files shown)


j9liu:
Thanks @kring !

PR #721, opened 2026-09-01
Fix Linux CI build failure

kring:
The failure was an error while running ezvcpkg during the configure step:

```
"/home/runner/.ezvcpkg/56bb2411609227288b70117ead2c47585ba07713.lock"
  creation failed (check permissions).
```

Why would we be unable to write that file? One possibility was that the runner was out of disk space, but that seemed unlikely. The runner has plenty of disk space, and Unity builds don't need an excessive amount like Unreal builds do. Plus it would be quite a coincidence to consistently run out just as we try to write that tiny file.

So I think maybe it truly is a permission problem. The runner isn't allowing us to write to a dot directory under the home directory perhaps? I tried to find something in the runner or GitHub Actions changelog about that, but came up empty. But lacking a better idea, I tried changing the ezvcpkg root to `~/ezvcpkg` instead of `~/.ezvcpkg`, and that seems to have worked.

Or there was some unrelated problem in GitHub's infrastructure that just happened to be fixed just as I submitted this... 🤷

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -350,10 +350,16 @@ jobs:
         uses: actions/setup-dotnet@v4
         with:
           dotnet-version: "10.0.103"
+      - name: Debug build problem
+        run: |
+          ls -la /home/runner
+          df -h
+          echo "EZVCPKG_BASEDIR=/home/runner/ezvcpkg" >> $GITHUB_ENV
+          mkdir ~/ezvcpkg
       - name: Cache vcpkg artifacts
         uses: actions/cache@v4
         with:
-          path: "~/.ezvcpkg"
+          path: "~/ezvcpkg"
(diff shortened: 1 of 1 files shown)


kring:
Hmm another possibility is that something was wrong with the cached ezvcpkg directory. Bad permissions or something. In which case this PR fixes it incidentally by invalidating the cache.

kring:
I re-ran a main build. It still failed, so it doesn't appear to have just been a temporary GitHub problem. I also noticed that the ezvcpkg cache was not found in that main build:

> Cache not found for input keys: vcpkg-linux-alma-45b76b2ad38ccc2bd7fbeea79358b12f95d802c15172714251915c0696c75df4-e34040f7333c658aec54a9b2420ba45c08bfc68d0541087eca411be964c48986, vcpkg-linux-alma-45b76b2ad38ccc2bd7fbeea79358b12f95d802c15172714251915c0696c75df4-

So my theory above about a something sketchy in the cache can't be right, either.

Which leaves us with the original explanation, even if it it seems unlikely: the runner isn't letting us write to:
`/home/runner/.ezvcpkg/56bb2411609227288b70117ead2c47585ba07713.lock`
but is happy for us to write to:
`/home/runner/ezvcpkg/56bb2411609227288b70117ead2c47585ba07713.lock`

Well, one other explanation I'm thinking of just now: maybe my explicit creation of `/home/runner/ezvcpkg` in this branch is what is enabling it to work, not the new directory name.

